PR #29, opened 2014-06-02
Added support for non-camel-case variable code conventions

yannickl88:
Not all people use the camel case format for variable. A popular code style is to use underscores in the variable names like ```variable_name```, but use a getter like ```getVariableName```. This is however not suported by the ClosureExpressionVisitor since it will just check for an accessor concatinated with the fieldname.

This adds an extra check so that this code style is also supported by the ClosureExpressionVisitor.


iltar:
As it won't be merged in the current state, I propose the following to remain fully BC with camel-casing as default:

| Properties available        | Expected method call | Gets property 
|-----------------------------|----------------------|--------------
| `$foo`                      | `getFoo()`           | ~
| `$foo_bar`                  | `getFooBar()`        | ~
| `$foo_bar_baz`              | `getFooBarBaz()`     | ~ 
| `$fooBar`                   | `getFooBar()`        | ~
| `$fooBarBaz`                | `getFooBarBaz()`     | ~
| `$foo_bar`, `$fooBar`       | `getFooBar()`        | `$fooBar`
| `$foo_bar_baz`,`$fooBarBaz` | `getFooBarBaz()`     | `$fooBarBaz`

Another solution could be to make it configurable on how to decide what the method name should be. This could be done by creating an interface and create a default implementation for both. This might be more flexible for the future.

rickprent:
+1

haroldiedema:
+1!

Any news on this matter?

yvoyer:
I made a fix (#62) to enable the configuration of a method name when calling `ArrayCollection::matching()` 

    Criteria::create()
        ->andWhere(Criteria::expr()->eq('yourWeirdMethodName', true));

@iltar @yannickl88 @Ocramius Would it solve your use case?

yannickl88:
Sadly no, your fix means that I still have to implement an extra method (with the exact same name as the field). And this format is not conforming to our code style.

yvoyer:
@yannickl88 The method name do not need to be the same as the field. it can be whatever you want in any format, as long as the method do not take any parameters.

You are right that you'll need a new method, but the advantage is that it does not need to follow the standard of Doctrine (`get` or `is`).

ie.

    class Foo 
    {
        private $bar = 12;

        public function i_can_ChooseTheFormat() 
        {
            return $this->bar; // or some other logic 
        }
    }

    $criteria = Criteria::create()
        ->andWhere(Criteria::expr()->eq('i_can_ChooseTheFormat', 12));

    $collection = new ArrayCollection(array(new Foo());
    $collection->matching($criteria); // Would return the Foo object

yannickl88:
Your usecase is for ArrayCollections. We are working with doctrine entities and persistent collections. So when you do a matching criteria on a non-hydrated collection with will actually create DQL statements to fetch a subset. Which in your case would result us querying a non-existent property.

Your example would thus result (for non-hydrated collections) in the following DQL:
```
SELECT f
FROM Foo
WHERE f.getBarBaz = 'foo'
```
However, this will result in an error since the getBarBaz is not a defined or mapped property

hboomsma:
Any updates? or other ways you would like to see it implemented, because this issue keeps on bugging me.

nicoschoenmaker:
:+1:

haroldiedema:
Over 1,5 years later...any progress on this, @Ocramius ?
We're still having difficulties with this.

Thanks in advance.

mikeSimonson:
Closing as #57 got merged.

PR #35, opened 2014-07-17
Added addUnique method to Collection

mmoreram:
- Elements are only added if they do not exist in the collection yet


JarJak:
Will methods like addUnique, unique be implemented in 3.0?

PR #37, opened 2014-08-05
 Offer single entry point to add / set elements

ronan-gloo:
To simplify implementation of specific behavior (i.e: validating data) when extending ArrayCollection, a single entry point for all "set" operations would be welcome.
The main downside of this PR is the operation made in constructor: we need to manually loop through elements in order to add them one by one. This have a significant performance impact on very large collections.


Ocramius:
Closing as "won't fix"

PR #45, opened 2014-10-25
Improve Comparison::CONTAINS: allow to use the SQL wildcard characters

Fedik:
This pull additional to doctrine/doctrine2#1150
And allow to use the [SQL wildcard characters](http://www.w3schools.com/sql/sql_wildcards.asp) `%` and `_`  for  `Comparison::CONTAINS`

Allow to build the `contains` expressions like:

``` php
Criteria::expr()->contains('myField', 'some string%');
Criteria::expr()->contains('myField', '%some string');
Criteria::expr()->contains('myField', '%10\%'); // search for 10% at end of the string
Criteria::expr()->contains('myField', 's_m_ string');
```

Note:  this changes can make b/c issue for existing `Comparison::CONTAINS` expressions, when someone use not escaped `%`, `_` .

### DIFF
--- a/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
+++ b/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
@@ -152,7 +152,28 @@ class ClosureExpressionVisitor extends ExpressionVisitor
 
             case Comparison::CONTAINS:
                 return function ($object) use ($field, $value) {
-                    return false !== strpos(ClosureExpressionVisitor::getObjectFieldValue($object, $field), $value);
+                    $field_value = ClosureExpressionVisitor::getObjectFieldValue($object, $field);
+
+                    // Check whether we have a wildcard characters, and build the regular expression
+                    $pattern = str_replace('\%', 'SQLWILDCARDESCAPEDMANY', $value);
+                    $pattern = str_replace('\_', 'SQLWILDCARDESCAPEDONE', $pattern);
+                    if(strpos($pattern, '%') !== false || strpos($pattern, '_') !== false) {
+                        // Build regexp
+                        $pattern = preg_quote($pattern);
+                        $pattern = str_replace('%', '.*', $pattern);
+                        $pattern = str_replace('_', '.{1}', $pattern);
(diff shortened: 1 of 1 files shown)


Ocramius:
This is probably an actual `like`/`ilike` operator to be introduced

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -152,7 +152,28 @@ public function walkComparison(Comparison $comparison)
 
             case Comparison::CONTAINS:
                 return function ($object) use ($field, $value) {
-                    return false !== strpos(ClosureExpressionVisitor::getObjectFieldValue($object, $field), $value);
+                    $field_value = ClosureExpressionVisitor::getObjectFieldValue($object, $field);
```
Ocramius: This API is uncovered by tests

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -152,7 +152,28 @@ public function walkComparison(Comparison $comparison)
...
+
+                    // Check whether we have a wildcard characters, and build the regular expression
+                    $pattern = str_replace('\%', 'SQLWILDCARDESCAPEDMANY', $value);
```
Ocramius: These strings should probably be class constants

Fedik:
@Ocramius I think if like/notlike #50  will be accepted then current issue can be closed ... 
I keep it open as get not to much feedback about like/notlike

stof:
The current issue **cannot** be closed even if #50 is accepted. this PR is about fixing a **bug** in the current implementation, where the SQL generated for CONTAINS is not consistent with the behavior of Doctrine Common.

Ocramius:
Thanks for clearing that up, @stof - missed that part completely while
responding to the issue today.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

Fedik:
@stof  I thought doctrine/doctrine2#1150  about SQL .. (where you told about unescaped value)

PR #46, opened 2014-10-26
Added prev to collections

wshafer:
Added prev() to array collections.  This is needed when you want to cycle through the array in reverse order.


Ocramius:
Can't fix this, as it is a bc break

PR #49, opened 2014-11-30
Keep keys when using array collection's matching()

marek-saji:
Use `uksort()` instead of `usort()`.


Ocramius:
Manually merged after adding tests, thanks!

PR #50, opened 2014-12-10
Like or Not Like - that's where the question

Fedik:
New `Comparison::LIKE` and `Comparison::NOTLIKE` as alternative solution for #45 :dancer: 

Support next wildcard characters:
`%` - any character(s), any amount
`_` - any character, only one
`\%` - escape for `%`
`\_` - escape for `_`


lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -155,6 +155,37 @@ public function walkComparison(Comparison $comparison)
...
+                    $pattern = str_replace('\_', 'SQLWILDCARDESCAPEDONE', $pattern);
+
+                    // Check whether we have a wildcard characters, and build the regular expression
+                    if(strpos($pattern, '%') !== false || strpos($pattern, '_') !== false) {
+                        // Build regexp
+                        $pattern = preg_quote($pattern);
+                        $pattern = str_replace('%', '.*', $pattern);
+                        $pattern = str_replace('_', '.{1}', $pattern);
+                        $pattern = str_replace('SQLWILDCARDESCAPEDMANY', '\\%', $pattern);
+                        $pattern = str_replace('SQLWILDCARDESCAPEDONE', '\\_', $pattern);
```
stof: this implementation is flawed: it will break for strings containing `SQLWILDCARDESCAPEDONE` already.

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -155,6 +155,37 @@ public function walkComparison(Comparison $comparison)
                     return false !== strpos(ClosureExpressionVisitor::getObjectFieldValue($object, $field), $value);
                 };
 
+            case Comparison::LIKE:
+            case Comparison::NOTLIKE:
+                $like = $comparison->getOperator() === Comparison::LIKE ? true : false;
```
stof: `$like = $comparison->getOperator() === Comparison::LIKE;` is enough. No need to use a ternary operator to turn a boolean into itself.

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -155,6 +155,37 @@ public function walkComparison(Comparison $comparison)
...
+
+                    // Check whether we have a wildcard characters, and build the regular expression
+                    if(strpos($pattern, '%') !== false || strpos($pattern, '_') !== false) {
+                        // Build regexp
+                        $pattern = preg_quote($pattern);
+                        $pattern = str_replace('%', '.*', $pattern);
+                        $pattern = str_replace('_', '.{1}', $pattern);
+                        $pattern = str_replace('SQLWILDCARDESCAPEDMANY', '\\%', $pattern);
+                        $pattern = str_replace('SQLWILDCARDESCAPEDONE', '\\_', $pattern);
+                        $pattern = '/^' . $pattern . '$/i';
```
stof: All the pattern building should be done outside the closure actually, to avoid rebuilding it for each object checked in the collection

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -155,6 +155,37 @@ public function walkComparison(Comparison $comparison)
                     return false !== strpos(ClosureExpressionVisitor::getObjectFieldValue($object, $field), $value);
                 };
 
+            case Comparison::LIKE:
+            case Comparison::NOTLIKE:
+                $like = $comparison->getOperator() === Comparison::LIKE ? true : false;
+                return function ($object) use ($field, $value, $like) {
+                    $field_value = ClosureExpressionVisitor::getObjectFieldValue($object, $field);
```
stof: the variable name should be camelCased

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -155,6 +155,37 @@ public function walkComparison(Comparison $comparison)
...
+                    if(strpos($pattern, '%') !== false || strpos($pattern, '_') !== false) {
+                        // Build regexp
+                        $pattern = preg_quote($pattern);
+                        $pattern = str_replace('%', '.*', $pattern);
+                        $pattern = str_replace('_', '.{1}', $pattern);
+                        $pattern = str_replace('SQLWILDCARDESCAPEDMANY', '\\%', $pattern);
+                        $pattern = str_replace('SQLWILDCARDESCAPEDONE', '\\_', $pattern);
+                        $pattern = '/^' . $pattern . '$/i';
+
+                        return (bool) ($like ? preg_match_all($pattern, $field_value) : !preg_match_all($pattern, $field_value));
```
stof: wouldn't preg_match be enough rather than preg_match_all ? We don't need to start matching it once we have found a match

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
stof: And the regex should not be case insensitive. It is inconsistent with the case of no wildcard being used.

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -155,6 +155,37 @@ public function walkComparison(Comparison $comparison)
...
+                    $field_value = ClosureExpressionVisitor::getObjectFieldValue($object, $field);
+
+                    // Replace the escaped characters to placeholder
+                    $pattern = str_replace('\%', 'SQLWILDCARDESCAPEDMANY', $value);
+                    $pattern = str_replace('\_', 'SQLWILDCARDESCAPEDONE', $pattern);
+
+                    // Check whether we have a wildcard characters, and build the regular expression
+                    if(strpos($pattern, '%') !== false || strpos($pattern, '_') !== false) {
+                        // Build regexp
+                        $pattern = preg_quote($pattern);
```
stof: you need to use `preg_quote($pattern, '/');` to be sure that `/` is escaped as well given you use it as delimiter

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
Fedik: true, but what can be alternative? 
use `DOCTRINECOLLECTIONSSQLWILDCARDESCAPEDMANY` ? :wink:

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
Fedik: you are right .. just to much work today :smile:

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
Fedik: but SQL `LIKE` case insensitive or?

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
Fedik: it work different for multiline string, if I right remember

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
Fedik: good tip

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
Fedik: hm, ok will do  case sensitive ... in SQL it depend from the database

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
stof: not if you don't add the `/m` modifier

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
stof: in PostgreSQL, it is case sensitive. in MySQL, it depends on the field collation (case insensitive collactions will match in a case insensitive way)

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -155,6 +155,37 @@ public function walkComparison(Comparison $comparison)
...
+
+                    // Check whether we have a wildcard characters, and build the regular expression
+                    if(strpos($pattern, '%') !== false || strpos($pattern, '_') !== false) {
+                        // Build regexp
+                        $pattern = preg_quote($pattern, '/');
+                        $pattern = str_replace('%', '.*', $pattern);
+                        $pattern = str_replace('_', '.{1}', $pattern);
+                        $pattern = str_replace('SQLWILDCARDESCAPEDMANY', '\\%', $pattern);
+                        $pattern = str_replace('SQLWILDCARDESCAPEDONE', '\\_', $pattern);
+                        $pattern = '/^' . $pattern . '$/';
```
stof: Because of your other changes, some of my comments disappeared without being fixed.
All the pattern building should be done outside the closure given that it does not depend on the object. This will avoid rebuilding it for each object of the collection

Fedik: disappeared, but I not forgot :wink:

if I place it like:

``` php
$like = $comparison->getOperator() === Comparison::LIKE;
/// here the pattern building
return function ($object) use ($field, $value, $like) {
```

it will be good?
just not sure

stof: yes it will

Fedik: hope now is better :smile:

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
Fedik: you are right, I changed it

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -155,6 +155,42 @@ public function walkComparison(Comparison $comparison)
...
+                    $value = str_replace('\%', '%', $value);
+                    $value = str_replace('\_', '_', $value);
+                }
+
+                return function ($object) use ($field, $value, $like, $pattern) {
+                    $fieldValue = ClosureExpressionVisitor::getObjectFieldValue($object, $field);
+
+                    if($pattern) {
+                        return (bool) ($like ? preg_match($pattern, $fieldValue) : !preg_match($pattern, $fieldValue));
+                    }
```
stof: instead of checking whether a preg_match or an equality should be used each time the function is called, you could check it outside, returning different functions

Fedik: yes, good idea

Fedik:
hm, hm, I wondered why it marked :bug: ?  I thought it is "new feature" :smile:

Fedik:
pr is updated for 2.x

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
Fedik: This is fixed

PR #51, opened 2015-01-19
[WIP] Added AbstractWrappedCollection support.

guilhermeblanco:


lib/Doctrine/Common/Collections/AbstractWrappedCollection.php
```diff
@@ -0,0 +1,276 @@
...
+
+use Closure;
+
+/**
+ * An AbstractWrappedCollection is a Collection implementation that wraps another Collection.
+ *
+ * @since  1.3
+ * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
+ */
+abstract class AbstractWrappedCollection extends ArrayCollection
```
deeky666: Why extending `ArrayCollection`? Isn't implementing `Collection` interface directly a better choice? No need to add a dependency on `ArrayCollection` here IMO as your contructor isn't tied to `ArrayCollection` either.

lib/Doctrine/Common/Collections/AbstractWrappedCollection.php
```diff
@@ -0,0 +1,276 @@
...
+    protected $collection;
+
+    /**
+     * Initializes a new ArrayCollection.
+     *
+     * @param array $elements
+     */
+    public function __construct(Collection $collection = null)
+    {
+        $this->collection = $collection ?: new ArrayCollection();
```
deeky666: I think I would not make a default instance here, that is too much implementation detail of the abstract collection wrapper IMO and should rather be a decision to be made in the concrete implementation.
I would suggest making `$collection` a required parameter instead.

lib/Doctrine/Common/Collections/AbstractWrappedCollection.php
```diff
@@ -0,0 +1,276 @@
...
+{
+    /**
+     * @var Collection
+     */
+    protected $collection;
+
+    /**
+     * Initializes a new ArrayCollection.
+     *
+     * @param array $elements
```
deeky666: Please fix the DocBlock param and description here.

Ocramius:
The naming does not express the intent here: why is this needed? Where is the change of behavior, compared to a normal collection?

beberlei:
Again here: Why? You need to explain why we need this :)

lib/Doctrine/Common/Collections/AbstractWrappedCollection.php
guilhermeblanco: Makes sense... will address this

guilhermeblanco:
@Ocramius @beberlei It's a Collection that wraps another one.
Right now, PersistentCollection already does that by wrapping a Collection of elements.
I'm implementing this to:
# Ditch lines of code from PersistentCollection in ORM
# Add support for custom collections which is currently WIP here: https://github.com/doctrine/doctrine2/compare/custom-collections

lib/Doctrine/Common/Collections/AbstractWrappedCollection.php
guilhermeblanco: I can't... here's the final idea:

```php
/**
 * @ORM\Entity
 */
class User
{
    private $roles;

    public function __construct()
    {
        $this->roles = new RoleCollection();
    }
}
```

danizord:
I think `AbstractCollectionWrapper` is a more clear naming

stof:
@guilhermeblanco how does this compare to the current AbstractLazyCollection ?

guilhermeblanco:
@stof There're 2 main differences:
* No initialization
* `partition()` must return instances of static

stof:
@guilhermeblanco a PersistentCollection is actually a lazy one currently. It has initialization logic

guilhermeblanco:
@stof I realized that after you pointed out... will update on ORM branch later today.
It still does not invalidate the purpose of this class.

Ocramius:
I don't see a correlation between `AbstractCollectionWrapper` and the functionality being provided though: really needs something more explicit/clear.

danizord:
Well, this is not wrapped, but it wraps.

Ocramius:
> it wraps.

yes, but what does it **do**?

Ocramius:
Closing here - no longer needed.

PR #52, opened 2015-01-20
Made initialized protected for extensibility.

guilhermeblanco:
I'm working on making PersistentCollection extends AbstractLazyCollection and it is required to have access to initialized property. This PR allows that.

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -37,9 +37,9 @@ abstract class AbstractLazyCollection implements Collection
     protected $collection;
 
     /**
-     * @var bool
+     * @var boolean
      */
-    private $initialized = false;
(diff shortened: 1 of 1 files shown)


stof:
should it be protected, or should it have only a protected getter to ensure that mutating it is done only in initialize ?

Ocramius:
I'd rather have a getter here, so that we can refactor later on. This also allows composition rather than forcing direct inheritance.

Yes, I know it's an abstract type.

beberlei:
For performance reasons i prefer the `protected` variable instead of a getter tbh.

guilhermeblanco:
@Ocramius we already a getter `isInitialized()`. I'd require a setter.
PersistentCollection in ORM calls initialized around 10 times, which I'd prefer to have as protected instead of a method for performance reasons too.

fprochazka:
Entity might have it's own `$initialized` property. This would break it imho.

PR #53, opened 2015-02-04
Added possibility to ignore case for expressions

erikverheij:


mikeSimonson:
@erikverheij Seems like overkill. Closing

PR #54, opened 2015-03-11
Add reduce method, add unit test

nike-17:
All attempts to use http://php.net/manual/en/function.array-reduce.php leads to warning
Warning: array_reduce() expects parameter 1 to be array, object given
There is some workaround about it, but as we have map method in collection it make sense to have also reduce.
What was done
- Added reduce method
- Added unit test to CollectionTest

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -242,6 +242,15 @@ abstract class AbstractLazyCollection implements Collection
     /**
      * {@inheritDoc}
      */
+    public function reduce(Closure $func, $initial = null)
+    {
+        $this->initialize();
+        return $this->collection->reduce($func, $initial);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     public function partition(Closure $p)
     {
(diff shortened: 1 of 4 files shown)


stof:
Adding a method in the interface is a BC break: it breaks all other implementations of the interface (and having implementations in other packages is a totally supported use case, given that the ORM and all ODM projects are doing it for instance)

nike-17:
Yes. We also have to update PersistentCollection and some other implementation.
But to have map without reduce simce like not complete implementation to me.
As an option, we can as well extract both map and reduce to separate interface

Ocramius:
The collection interface will unlikely ever be "complete".

Sadly, I don't think a separate interface would be useful at all as-is.

I'd rather defer this sort of API study to libraries such as
https://github.com/morrisonlevi/SPL-Collections

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

guiwoda:
@Ocramius Will this be merged, or your response actually means that Doctrine collections will stay as-is forever? (or until a major version bump?)

Ocramius:
We can't add API in 2.x due to BC concerns

PR #55, opened 2015-04-03
Comparison startsWith, endsWith

pine3ree:
Trying to propose again a couple of common comparison test (related to contains test).
I also have the related PR for doctrine2 repo ready.

### DIFF
--- a/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
+++ b/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
@@ -155,6 +155,16 @@ class ClosureExpressionVisitor extends ExpressionVisitor
                     return false !== strpos(ClosureExpressionVisitor::getObjectFieldValue($object, $field), $value);
                 };
 
+            case Comparison::STARTS_WITH:
+                return function ($object) use ($field, $value) {
+                    return 0 === strpos(ClosureExpressionVisitor::getObjectFieldValue($object, $field), $value);
+                };
+
+            case Comparison::ENDS_WITH:
+                return function ($object) use ($field, $value) {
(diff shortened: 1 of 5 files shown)


beberlei:
For reference, we have contains already, which does a string match on any part of the string.

mikeSimonson:
@beberlei This seems good to go to me.

mikeSimonson:
merged in https://github.com/doctrine/collections/pull/78

stof:
Have you sent a PR updating the ORM to support these new operators ?

mikeSimonson:
@stof Against what code that PR need to be made ?

stof:
@mikeSimonson the ORM code turning expression into SQL queries for usage in persistent collections: https://github.com/doctrine/doctrine2/blob/c1943624ab1260c629316bab104dc5130c060154/lib/Doctrine/ORM/Persisters/SqlExpressionVisitor.php

mikeSimonson:
Those are not compositeExpression but comparison ( the method above this one).
It doesn't look like any change is needed there but maybe there https://github.com/doctrine/doctrine2/blob/c1943624ab1260c629316bab104dc5130c060154/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L99 ?

stof:
hmm yeah, I forgot that the SqlExpressionVisitor actually delegates to each persister to build the comparison.

grizzm0:
Any ETA on the ORM support? ;)

I'm not that into the ORM source. Would it be enough to just add them to the comparisonMap?

Ocramius:
@grizzm0 nobody made a patch for this, as far as I know. `SqlExpressionVisitor` is where you want to start :-)

grizzm0:
@Ocramius Probably deeper than my understanding of the ORM. :/

PR #57, opened 2015-05-13
non-camel-case variable code conventions (with extra tests)

iltar:
Because people keep running into issues with collections and `$underscore_properties` , I have added extra tests on the changes of #29 to to prove that it's backwards compatible. It seems like that one won't be merged because the test cases are missing.

This PR includes the exact same changes as #29 but with 3 extra tests.

### DIFF
--- a/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
+++ b/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
@@ -58,18 +58,6 @@ class ClosureExpressionVisitor extends ExpressionVisitor
             return $object->$accessor();
         }
 
-        // camelcase field name to support different variable naming conventions
-        $ccField = preg_replace_callback('/_(.?)/', function($matches) { return strtoupper($matches[1]); }, $field);
-
-        foreach ($accessors as $accessor) {
-            $accessor .= $ccField;
-
-            if ( ! method_exists($object, $accessor)) {
-                continue;
-            }
-
-            return $object->$accessor();
-        }
 
         // __call should be triggered for get.
         $accessor = $accessors[0] . $field;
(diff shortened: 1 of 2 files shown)


iltar:
Failing tests are only on `hhvm-nightly`, which is not supported anymore according to the error message.

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -45,12 +45,10 @@ public static function getObjectFieldValue($object, $field)
         if (is_array($object)) {
             return $object[$field];
         }
-
```
stof: please don't remove these empty lines

stof:
the HHVM-nightly builds are now removed

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
iltar: Ah my bad, I was debugging a bit in a particular test case, I will revert this on Monday.

TomasVotruba:
What needs to be done here?

iltar:
I actually forgot to provide the last patch, which is added now. 

@Ocramius is there anything this PR is still missing in order to get the fix through?

mikeSimonson:
@iltar Baswed on the discution on #29 I am going to consider that this implement the desired behaviour.

iltar:
@mikeSimonson You have our thanks!

yannickl88:
@mikeSimonson Do you have some info when there will be a release that includes this pull?

mikeSimonson:
I need to fix a few stuff that I pulled here before to release. I am not making any promise but I hope soon.

yannickl88:
okay cool! looking forward to it

maresja1:
This is not enough. Suppose my field is called `isCompany`, as it is `bool` and name `company` would suggest a reference to a company. I really don't want to call the getter method for it `isIsCompany` or `getIsCompany` - I consider that as ugly. Why can't I call the getter same as the field. I suggest adding possibility for users of Doctrine to override the naming convention, or at least consider naming standard where getters are called the same as the fields.

yannickl88:
@maresja1 What you are suggesting would be a great addition. However, this would require an extension point for the `ClosureExpressionVisitor`. I would recommend creating a new Issue since this one is already merged.

PR #60, opened 2015-06-25
Enable accessor for underscore naming strategy

bradhanebury:
A small fix for those of us forced to use projects with the underscore naming strategy and want to use the Criteria class.


TomasVotruba:
@bradhanebury Could you provide tests?

Ref #57

mikeSimonson:
@bradhanebury Is your use case fixed by #57 ?

bradhanebury:
Appears so!

PR #61, opened 2015-06-30
Add a new SetCollection

MisatoTremor:
As Travis fails due to internal errors: [Build Status](https://travis-ci.org/MisatoTremor/collections)

### DIFF
--- a/lib/Doctrine/Common/Collections/SetCollection.php
+++ b/lib/Doctrine/Common/Collections/SetCollection.php
@@ -1,0 +1,65 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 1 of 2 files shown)


beberlei:
This implementation violates the liskov substitution principle, because the following is possible but makes no sense:

     $set1["foo"] = "bar";
     $set2["bar"] = "bar";

MisatoTremor:
Why does adding a element to 2 different sets with different keys make no sense? Maybe you meant something else ;)

However, you are correct that I violated Liskov, which was done to avoid heavy repetition and I thought that it would not be a real issue in practice.
Another solution would be an abstract "base" class to avoid both violations (but adding another layer of inheritance). What do you think about that?

mikeSimonson:
@MisatoTremor In the context of a Set I don't understand the presence of a set(key, value) method.
There is no notion of key in a set.

MisatoTremor:
@mikeSimonson Of course you're right about that.
But afaik all collections must implement the Collection interface and changing the signature of the set method would violate PHP strict standards.

mikeSimonson:
To me it suggest that it shouldn't extends ArrayCollection

MisatoTremor:
Sorry, my explanation might have been a bit unclear.
Giving the SetCollection it's complete own implementation would not change the fact that the method is in the Collection interface: ``public function set($key, $value);`` Therefor it must be implemented with a matching signature and that includes a key.

mikeSimonson:
That's what I mean, in this case I don't think it's a good idea to try to implement a set because you can't do it in a clean way.

MisatoTremor:
OK, good point then.
Solving this problem would involve generalizing / splitting the interface which, given the huge BC break this would be, won't happen anytime soon, if at all.

Ocramius:
Closing due to https://github.com/doctrine/collections/pull/61#issuecomment-132718623

PR #63, opened 2015-08-04
Fix the branch alias for master

stof:
1.3.0 is already released, so the master branch cannot be 1.2.x-dev

I bumped it to 1.3.x as the branch does not currently contain any new feature compared to 1.3.0, so the next release could be 1.3.1 or 1.4.0 and I don't know it yet (it depends of what will happen)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.2.x-dev"
+            "dev-master": "1.3.x-dev"
         }
     }
 }


Ocramius:
@stof should it be 1.4.x-dev then?

stof:
well, the master branch does not yet contain any feature mandating its version to become 1.4 according to semver (the only change compared to 1.3.0 is an improvement in the CI setup). So for now, it is really a 1.3.x-dev.

It will make sense to bump it to 1.4.x-dev when we have new features in the branch (which means we will know that the unreleased code will become 1.4.0)

Ocramius:
Alright, :+1:

PR #65, opened 2015-08-19
Upd ArrayCollection.php (implement JsonSerializable)

NorthV:
ArrayCollection can be encoded with json_encode() after implementation of JsonSerializable.
(PHP 5 >= 5.4.0, PHP 7)
http://fi2.php.net/manual/en/class.jsonserializable.php

### DIFF
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -22,6 +22,7 @@ namespace Doctrine\Common\Collections;
 use ArrayIterator;
 use Closure;
 use Doctrine\Common\Collections\Expr\ClosureExpressionVisitor;
+use JsonSerializable;
 
 /**
  * An ArrayCollection is a Collection implementation that wraps a regular PHP array.
@@ -36,7 +37,7 @@ use Doctrine\Common\Collections\Expr\ClosureExpressionVisitor;
  * @author Jonathan Wage <jonwage@gmail.com>
  * @author Roman Borschel <roman@code-factory.org>
  */
-class ArrayCollection implements Collection, Selectable
+class ArrayCollection implements Collection, Selectable, JsonSerializable
 {
     /**
      * An array containing the entries of this collection.
@@ -61,6 +62,16 @@ class ArrayCollection implements Collection, Selectable
     public function toArray()
     {
         return $this->elements;
+    }
+
+    /**
+     * Returns an array representation of this object.
+     *
+     * @return array
+     */
+    public function jsonSerialize()
+    {
+        return $this->toArray();
     }
 
     /**


Ocramius:
This would lead to infinite recursion whenever you try to `json_serialize()` a bi-directional many-to-many association: seems like a real bad idea (plus it is useless unless we change the interface, which is not supported in 2.x)

NorthV:
This would not lead to a recursion itself.
Because by default if we try to json_encode($doctrineEntity) we will get result: `string(2) "{}"`.
Instead, we will have an opportunity to have standard json representation of array from the box. 
In this case, 
`json_encode(array(1,2,3, new DoctrineEntity())) `
and 
`json_encode(new ArrayCollection(array(1,2,3, new DoctrineEntity())))` 
will give us the exact same result: `[1,2,3,{}]`

Ocramius:
```php
$foo = new stdClass;
$bar = new stdClass;

$foo->bar = $bar;
$bar->foo = $foo;

echo json_encode($foo);
```

And that's how you make it [crash and burn](https://3v4l.org/5rVEA)

NorthV:
Thx for this example. But it is about `json_encode()` itself. 
Those who use `json_encode()` knows about this problem. And this problem is not the real reason to not to use `json_encode()` in some cases, when you deal this scalar data and arrays.

I think the ArrayCollection is extended array, thats why it would be great if it would be `json_encoded()` exactly like array.

```php
<?php

class ArrayCollection
{
    private $_elements;

    public function __construct(array $elements = array())
    {
        $this->_elements = $elements;
    }
}

class ArrayCollectionI implements \JsonSerializable
{
    private $_elements;

    public function __construct(array $elements = array())
    {
        $this->_elements = $elements;
    }

    public function toArray()
    {
        return $this->_elements;
    }

    /**
     * Returns an array representation of this object.
     *
     * @return array
     */
    public function jsonSerialize()
    {
        return $this->toArray();
    }
}

var_dump(json_encode(array(1,2,3)));
var_dump(json_encode(new ArrayCollection(array(1,2,3))));
var_dump(json_encode(new ArrayCollectionI(array(1,2,3))));
```
And example [make it](https://3v4l.org/XBcJt)
```
Output for 5.4.0 - 7.0.0beta3, hhvm-3.3.1 - 3.8.1
string(7) "[1,2,3]"
string(2) "{}"
string(7) "[1,2,3]"
```

NorthV:
And this how json_encode() aproach can be used to serialize DoctrineEntity 
http://stackoverflow.com/a/11325189/637210

Ocramius:
My point is: use a proper serializer, like JMSSerializer

NorthV:
I agree with your point. Proper serializers are great to use this complex objects like DoctrineEntity. 

But if we need to deal with scalar data or array i think usage of standard php functionality is more suitable, because of it's simplicity and performance. 
And the ArrayCollection can contain not only the DoctrineEntities, but scalar data hydrated to multidimensional array.

This implementation can reduse some code like this:
```php
if ($var instanceof ArrayCollection) {
   echo json_encode($var->toArray());
}
else {
   echo json_encode($var);
}
```
It's simple example, more complex is to handle multidimensional array with ArrayCollection in some array values.

PR #66, opened 2015-09-04
implement ExpressionBuilder->memberOf() method

ferjul17:
Hi,

The comparison MEMBER_OF was not implemented on Cretieria while it seems really easy to implement.
I implemented it and wrote the test.
I also tested it in my application.

### DIFF
--- a/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
+++ b/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
@@ -155,6 +155,11 @@ class ClosureExpressionVisitor extends ExpressionVisitor
                     return false !== strpos(ClosureExpressionVisitor::getObjectFieldValue($object, $field), $value);
                 };
 
+            case Comparison::MEMBER_OF:
+                return function ($object) use ($field, $value) {
+                    return in_array($value, ClosureExpressionVisitor::getObjectFieldValue($object, $field)->toArray());
+                };
+
             default:
                 throw new \RuntimeException("Unknown comparison operator: " . $comparison->getOperator());
         }
--- a/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
+++ b/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
@@ -158,12 +158,7 @@ class ClosureExpressionVisitor extends ExpressionVisitor
             case Comparison::MEMBER_OF:
                 return function ($object) use ($field, $value) {
                     $fieldValues = ClosureExpressionVisitor::getObjectFieldValue($object, $field);
-                    foreach ($fieldValues as $fieldValue) {
-                        if ($value == $fieldValue) {
-                            return true;
-                        }
-                    }
-                    return false;
+                    return in_array($value, is_array($fieldValues) ? $fieldValues : iterator_to_array($fieldValues));
                 };
 
             default:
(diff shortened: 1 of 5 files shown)


baileylo:
The parameter order seems a little odd to me. I'm assuming, the value parameter comes first to make the code more readable, `->memberOf('photography', 'hobbies')` could be read 'photography is a member of hobbies`. However, every other method has the reversed parameter order - `->memberOf('hobbies', 'photography')`.

Also, The code fails if the nested value is an array.

```php
$collection = new ArrayCollection([
    ['name' => 'Bill', 'hobbies' => ['biking', 'running', 'swimming']],
    ['name' => 'Todd', 'hobbies' => ['video games', 'laser tag']],
    ['name' => 'Casey', 'hobbies' => ['tennis', 'golf', 'photography']]
]);

$criteria = Criteria::create()->where(Criteria::expr()->memberOf('photography', 'hobbies'));

$collection->matching($criteria)->toArray();
```

ferjul17:
Thanks for your feedback.

As you said I inverted the parameters in order to make it more readable because in my opinion this is the most important thing. But changing the order of the parameter takes about 2s with my IDE.

I fixed my code. Unfortunatly, it is not as beautifull as the rest of the code cause I can not use the in_array function.

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -155,6 +155,17 @@ public function walkComparison(Comparison $comparison)
                     return false !== strpos(ClosureExpressionVisitor::getObjectFieldValue($object, $field), $value);
                 };
 
+            case Comparison::MEMBER_OF:
+                return function ($object) use ($field, $value) {
+                    $fieldValues = ClosureExpressionVisitor::getObjectFieldValue($object, $field);
+                    foreach ($fieldValues as $fieldValue) {
```
mikeSimonson: @ferjul17 Why can't you use in_array here ?

lib/Doctrine/Common/Collections/ExpressionBuilder.php
```diff
@@ -163,4 +163,15 @@ public function contains($field, $value)
...
         return new Comparison($field, Comparison::CONTAINS, new Value($value));
     }
+
+    /**
+     * @param string $field
+     * @param mixed  $value
+     *
+     * @return Comparison
+     */
+    public function memberOf($value, $field)
```
mikeSimonson: Can you please change the parameter order for consistency with all the other methods ?

ferjul17: I did it

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
ferjul17: @mikeSimonson ClosureExpressionVisitor::getObjectFieldValue may return an instance of Iterator which doesn't work with in_array.

mikeSimonson: Maybe using iterator_to_array and then in_arrau is a little bit cleaner ?

ferjul17: Would you prefer something like that ?

mikeSimonson:
@ferjul17 Thanks

ferjul17:
You're welcome :)

2016-03-27 19:00 GMT+02:00 mikeSimonson <notifications@github.com>:

PR #67, opened 2015-10-02
Inversion of logic expression - lower complexity

krzysiekpiasecki:


Ocramius:
@krzysiekpiasecki a couple of things:

 - return early in case of failures (as per object calisthenics), which actually simplifies code (you made it more complex, by nesting the entire success case)
 - `isset()` + `array_key_exists()` is actually a performance improvement hack, as `isset()` is not a function, and is compiled down to an opcode.

krzysiekpiasecki:
In my opinion you are wrong.

- There is 'and' operator between isset expression and array_key_exists in remove method - so it's not a hack - it's a crap.
- Use software metrics for complexity

Ocramius:
> There is 'and' operator between isset expression and array_key_exists in remove method - so it's not a hack - it's a crap.

boolean expressions are lazy by default, so the first expression evaluation should be the fastest ;-)

 > Use software metrics for complexity

The CCLOC is the same, as there is only one conditional

 > Use timer for performance

We actually do. The collections lib is so widespread and used in `O(n)` contexts that such an optimization is actually part of the design

krzysiekpiasecki:
How "and" operator can be lazy? :)

Ocramius:
It works also with "and". Anyway, the logic is just to handle the case when the element value is `null`

krzysiekpiasecki:
Yes, it's lazy when first expression is false. When is false? When element exists and it's not a null, so in most expected case :)

I was wrong.

PR #68, opened 2015-10-28
Update ClosureExpressionVisitor.php

DrSchimke:
Searching for elements within ArrayCollection using element's \DateTime attributes fails, since until now the comparison uses === instead of ==.

I would like to change the cases ::EQ and ::NEQ – is this ok?

(The matching functionality in PersistentCollection works fine, if I use EXTRA_LAZY...)

### DIFF
--- a/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
+++ b/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
@@ -112,12 +112,12 @@ class ClosureExpressionVisitor extends ExpressionVisitor
         switch ($comparison->getOperator()) {
             case Comparison::EQ:
                 return function ($object) use ($field, $value) {
-                    return ClosureExpressionVisitor::getObjectFieldValue($object, $field) === $value;
+                    return ClosureExpressionVisitor::getObjectFieldValue($object, $field) == $value;
                 };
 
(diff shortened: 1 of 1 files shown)


DrSchimke:
Here is an example for the problem:
```
$collection->add(['date' => new \DateTime('2015-01-01 00:00:00.000'), 'id' => 1]);
$collection->add(['date' => new \DateTime('2015-01-02 00:00:00.000'), 'id' => 2]);
var_dump($collection->matching(new Criteria(Criteria::expr()->eq('date', new \DateTime('2015-01-01 00:00:00.000'))))->toArray());
```

baileylo:
This is done by design and is documented in the [Doctrine doc's](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/basic-mapping.html#doctrine-mapping-types), you'll need to scroll down but it says:

> DateTime and Object types are compared by reference, not by value. Doctrine updates this values if the reference changes and therefore behaves as if these objects are immutable value objects.

stof:
@baileylo this is OK for updates, but not really for querying (DQL uses the datetime value when building queries)

mikeSimonson:
@stof If I understand you well, this PR should be merged right ?

beberlei:
This is not a good idea, comparing entities by reference can be a huge performance drain, because it happens recursively.

PR #71, opened 2016-03-22
Criteria factory

nikolaposa:
Factory which simplifies Criteria object creation, by enabling creation based on array specification.
### Usage

``` php
use Doctrine\Common\Collections\CriteriaFactory;

$factory = new CriteriaFactory();
$criteria = $factory->create(array(
    //(firstName = 'test' OR lastName = 'test') AND (status = 1 AND age >= 20)
    'expression' => array(
        array(
            '$or' => array(
                array('fld' => 'firstName', 'op' => '=', 'val' => 'test'),
                array('fld' => 'lastName', 'op' => '=', 'val' => 'test'),
            ),
        ),
        array(
            '$and' => array(
                array('fld' => 'status', 'op' => '=', 'val' => 1),
                array('fld' => 'age', 'op' => '>=', 'val' => 20),
            ),
        ),
    ),
    'orderings' => array('username' => 'ASC'),
    'max_results' => 10,
));
```
### Syntax

Criteria array specification should contain options supported by the Criteria object itself:
- `expression` - expression specification
- `orderings` - orderings array
- `first_result` - first result option
- `max_results` - max results option
#### Expressions

Expression can be a combination of composite and/or comparison expressions. Comparison must contain these properties:
- `fld` - comparison field
- `op` - comparison operator
- `value` - comparison value

Sequential array is an alternative for `$and`, so for example these two notations produce the same result:

``` php
array(
    '$and' => array(
        array('fld' => 'status', 'op' => '=', 'val' => 1),
        array('fld' => 'age', 'op' => '>=', 'val' => 20),
    )
)
```

and:

``` php
array(
    array(
        array('fld' => 'status', 'op' => '=', 'val' => 1),
        array('fld' => 'age', 'op' => '>=', 'val' => 20),
    )
)
```


lib/Doctrine/Common/Collections/Criteria.php
```diff
@@ -109,6 +109,12 @@ public function __construct(Expression $expression = null, array $orderings = nu
         }
     }
 
+    public static function fromArray(array $criteria) {
```
stof: I'm not sure we should have this static method

nikolaposa: @stof There is already `public static function create()`. `createFromArray()` maybe?

stof: this method was added to allow one-liner creation, as ``(new Criteria())->orderBy(array('foo' => 'DESC'))`` does not work on PHP 5.3, which is the min requirement of the library.

When building the whole criteria as a nested array externally and then converting it to a doctrine criteria, I don't expect wanting to continue changing it inline. And I would prefer keeping this criteria converter separate (I'm not sure I would want it in core, but I'm not the one making the decision anyway)

lib/Doctrine/Common/Collections/CriteriaFactory.php
```diff
@@ -0,0 +1,98 @@
...
+            && isset($expression['op'])
+            && isset($expression['val'])
+        ) {
+            return new Comparison($expression['fld'], $expression['op'], $expression['val']);
+        }
+
+        $expressionsArray = $expression;
+        $compositeExpressionType = CompositeExpression::TYPE_AND;
+
+        if (isset($expression['$or'])) {
```
stof: what happens when defining extra keys alongside ``$or`` or ``$and`` ?

stof: answer today: they are silently ignored

lib/Doctrine/Common/Collections/CriteriaFactory.php
```diff
@@ -0,0 +1,98 @@
...
+            throw Exception\InvalidCriteriaArrayException::fromInvalidExpression($expression);
+        }
+
+        if (
+            isset($expression['fld'])
+            && isset($expression['op'])
+            && isset($expression['val'])
+        ) {
+            return new Comparison($expression['fld'], $expression['op'], $expression['val']);
+        }
```
stof: Your current syntax is not helpful for typos:

```php
array(
    'fld' => 'username',
    'op' => '=',
    'val' => 'James',
)
```
will create a comparison expression, while the following (only 2 chars changed) will do weird things

```php
array(
    'fields' => 'username',
    'op' => '=',
    'val' => 'James',
)
```

You should better validate the structure of the array (and this may start by defining explicitly what is the syntax of your nested expression array)

lib/Doctrine/Common/Collections/CriteriaFactory.php
nikolaposa: There is some room for improvements here, I agree, I was even in a doubt whether `$or` and `$and` syntax is appropriate, but I couldn't come up with some better naming.

stof:
Please update the PR description with a description of the syntax of your criteria array (which is part of the feature you implement).
Btw, this syntax needs to be documented somewhere too.

nikolaposa:
@stof Updated PR description.

PR #72, opened 2016-03-23
Fix: Ignore composer.lock

localheinz:
This PR

* [x] adds `composer.lock` to `.gitignore`

:information_desk_person: It's apparently not checked in, so it should be ignored, right?


mikeSimonson:
@localheinz Thanks

PR #73, opened 2016-03-23
Fix: Assert expression is set when none has before

localheinz:
This PR

* [x] asserts that the argument of `andWhere()` and `orWhere` is set when an expression hasn't previously been set


mikeSimonson:
@localheinz Thanks

PR #80, opened 2016-04-07
Make the ArrayCollection extendable

Nek-:
Hello,

I wanted to extends the ArrayCollection of doctrine to add some cool features. But unfortunately the elements are private so I can't implement new methods. I felt too bad (to copy/paste the class entire class) for a so little word like `protected` missing :) .

Of course you can stay on position that you don't want doctrine collection to be inheritable, but if you want that I suggest to set this class final... But I don't get why it would be.

### DIFF
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -43,7 +43,7 @@ class ArrayCollection implements Collection, Selectable
      *
      * @var array
      */
-    private $elements;
+    protected $elements;
 
     /**
      * Initializes a new ArrayCollection.


Nek-:
By the way, I am extending the class to add `removeItemThat(Closure)` and things like that. Interested in a PR for that too ?

Ocramius:
You shouldn't extend ArrayCollection: use the Collection interface instead.

Nek-:
@Ocramius what about setting it final ?

But well, let's say I did it... by copy-pasting the whole class :') .

Ocramius:
> @Ocramius what about setting it final ?

Some internal classes actually extend `ArrayCollection` to retain type compatibility with userland usage, but indeed, we should make it `final` in a major version (not now though)

Nek-:
@Ocramius I still don't get why I must copy/paste the ArrayCollection while I could inherit from it. Any good reason to share so I can understand this choice ?

Ocramius:
@Nek- mostly because the class is not designed for inheritance in first place. If we did that, we'd have to guarantee BC on all of its internals, which we don't want to deal with in first place. C&P is perfectly valid here.

PR #82, opened 2016-04-22
Allow null as argument of ArrayCollection->filter() method

JarJak:
Allow null as argument of filter() method, same as array_filter does

### DIFF
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -295,7 +295,7 @@ class ArrayCollection implements Collection, Selectable
     /**
      * {@inheritDoc}
      */
-    public function filter(Closure $p)
+    public function filter(Closure $p = null)
     {
         return new static(array_filter($this->elements, $p));
     }


mikeSimonson:
@JarJak Thanks for the PR.

I don't think that  adding that weird feature of array_filter is a good idea.
You need to know that array filter will default to remove everything that is considered false and it's so much more expressive to just pass a callback that make that test.

I am personally against removing explicitness but I am not the one calling the shots so I will wait to see what others think about it.

Regardsssss

stof:
I'm also against this. Thus, saying you want the behavior of ``array_filter`` means that any implementation must implement this behavior even when it does not use ``array_filter`` internally. This is weird (btw, your PR does not make the interface describe what ``null`` does).

And btw, this is also a BC break, as it forces all implementators to make the argument optional and to accept ``null`` in their implementations. So it cannot be done anyway.

JarJak:
@mikeSimonson "it's so much more expressive to just pass a callback that make that test."

Interesting. Do you have more information about this topic?

how:
```
array_filter($array, function ($row) {
    return $row == false;
});
```

is faster than just:
`array_filter($array);`

Ocramius:
@JarJak he is talking about expressiveness, not performance. Passing an identity function to a filter method is much more clear than having `null` be equivalent to `$identity = function ($value) { return $value; }`

Ocramius:
Marked as `won't fix`. Indeed, the fact that `array_filter()` has a shorthand for identity functions shouldn't be made part of the collections API.

JarJak:
Ok thanks

PR #83, opened 2016-04-28
Spelling - comparision should be comparison

johnnypeck:
comparision should be comparison in the class docBlock.


mikeSimonson:
@johnnypeck Thanks

PR #84, opened 2016-05-13
Unmodifiable collection view

Sharom:
This pull-request adds class `UnmodifiableCollectionView` to be able to return a collection for read-only.

### DIFF
--- a/lib/Doctrine/Common/Collections/UnmodifiableCollectionView.php
+++ b/lib/Doctrine/Common/Collections/UnmodifiableCollectionView.php
@@ -1,0 +1,304 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
 vendor/
 composer.lock
 build/
+.idea
+*.iml
```
stof: -1 for that. these are not project-specific but system-specific. you should use a system-level ignore rule: https://help.github.com/articles/ignoring-files/

lib/Doctrine/Common/Collections/UnmodifiableCollectionView.php
```diff
@@ -0,0 +1,314 @@
...
+    {
+        return $this->collection->count();
+    }
+
+    /**
+     * @param string $message
+     *
+     * @return BadMethodCallException
+     */
+    private function createUnsupportedException($message)
```
stof: IMO, you don't need this private method

Sharom: agree

.gitignore
Sharom: fixed

lib/Doctrine/Common/Collections/UnmodifiableCollectionView.php
```diff
@@ -0,0 +1,304 @@
...
+use BadMethodCallException;
+
+/**
+ * Unmodifiable view of a collection.
+ *
+ * Wraps original collection and protects it from unplanned change.
+ * 
+ * @author Roman Shamritskiy <roman@litgroup.ru>
+ */
+class UnmodifiableCollectionView implements Collection
```
stof: would be great to implement Selectable too

Sharom: > would be great to implement Selectable too

I thought about it. But I still not sure that it's good solution.
What if a wrapped collection does not implement `Selectable`?
To throw `BadMethodCallException`? It's not clear.

Ocramius:
A collection that exposes not-implemented methods that throw exceptions seems wrong to me.

You now disallow methods that are expected to work (without failure) on the consumer side.

While I like where you are going with this, this is not the way to implement it. The correct implementation would be to segregate the read operations in own interfaces, and then implement only those.

I wouldn't do that though, as everyone currently relies on the `Collection` interface directly, so some more in-depth analysis of the problem and how we can evolve the library has to be done.

Closing as `invalid` due to the remarks above.

fprochazka:
@Sharom if you haven't already created a package, you might wanna look into this https://github.com/Kdyby/DoctrineCollectionsReadonly  :)

Sharom:
> The correct implementation would be to segregate the read operations in own interfaces, and then implement only those.

@Ocramius you are right. But... we already have a `Collection`. And I still want be able to return read-only collection from an entity, and prevent unexpected mutation of a collection from outside of aggregate.

Doctrine documentation contains one example:

```php
public function getReadComments() {
    return $this->commentsRead->toArray();
}
```

but this is wrong way. What about a lazy collection?

Implementation of  `UnmodifiableListView` is not what I invented yesterday (for example: https://api.dartlang.org/stable/1.16.0/dart-collection/UnmodifiableListView-class.html)

`BadMethodCallException` is a subtype of `RuntimeException` which evidences of breach of contract.
And this solution works quite successfully. You can think about immutability of a collection as about one of the states of a collection.

Segregation of interfaces for reading/changing of a collection may be a good solution, but it may be not. I didn't ever see implementation of generic collections with segregation of interfaces for reading and for modification.

`return $this->commentsRead->toArray()` – this is not a solution. If `UnmodifiableListView` not a solution too, let's think about necessity of a read-only collection together.

P.S. (If any sentences incorrect in the context of english, I am sorry about it.
I learn English ;))

Sharom:
@fprochazka

> if you haven't already created a package, you might wanna look into this https://github.com/Kdyby/DoctrineCollectionsReadonly :)

Thanks 👍

PR #86, opened 2016-05-27
Add unit tests for value and composite expression

tobiasoberrauch:
I added two unit tests. One for ValueExpression and the other for CompositeExpression.
The @covers annotation defines clearly which class I want to test (Subject under test)

I hope that this will fit to the standards :)
Cheers Tobias

### DIFF
--- a/tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
+++ b/tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
@@ -4,6 +4,7 @@ namespace Doctrine\Tests\Common\Collections\Expr;
 
 use Doctrine\Common\Collections\Expr\CompositeExpression;
 use Doctrine\Common\Collections\Expr\ExpressionVisitor;
+use Doctrine\Common\Collections\Expr\Value;
 use PHPUnit_Framework_TestCase as TestCase;
 
 /**
@@ -13,25 +14,45 @@ use PHPUnit_Framework_TestCase as TestCase;
 class CompositeExpressionTest extends TestCase
 {
     /**
-     * @return CompositeExpression
+     * @return array
      */
-    public function testConstructor()
+    public function invalidDataProvider()
     {
-        $type = CompositeExpression::TYPE_AND;
-        $expressions = array(
-            $this->getMock('Doctrine\Common\Collections\Expr\Expression'),
+        return array(
+            array(
+                'expression' => new Value('value'),
+            ),
+            array(
+                'expression' => 'wrong-type',
+            ),
         );
-
-        return new CompositeExpression($type, $expressions);
     }
 
     /**
-     * @depends testConstructor
+     * @dataProvider invalidDataProvider
      *
-     * @param CompositeExpression $compositeExpression The given expression.
+     * @param $expression
+     * @return void
      */
-    public function testGetType(CompositeExpression $compositeExpression)
+    public function testExceptions($expression)
     {
+        $this->setExpectedException('\RuntimeException');
+
+        $type = CompositeExpression::TYPE_AND;
+        $expressions = array(
+            $expression,
+        );
+
+        new CompositeExpression($type, $expressions);
+    }
+
+    /**
+     * @return void
+     */
+    public function testGetType()
+    {
+        $compositeExpression = $this->createCompositeExpression();
+
         $expectedType = CompositeExpression::TYPE_AND;
         $actualType = $compositeExpression->getType();
 
@@ -39,12 +60,27 @@ class CompositeExpressionTest extends TestCase
     }
 
     /**
-     * @depends testConstructor
-     *
-     * @param CompositeExpression $compositeExpression The given expression.
+     * @return CompositeExpression
      */
-    public function testGetExpressionList(CompositeExpression $compositeExpression)
+    protected function createCompositeExpression()
     {
+        $type = CompositeExpression::TYPE_AND;
+        $expressions = array(
+            $this->getMock('Doctrine\Common\Collections\Expr\Expression'),
+        );
+
+        $compositeExpression = new CompositeExpression($type, $expressions);
+
+        return $compositeExpression;
+    }
+
+    /**
+     * @return void
+     */
+    public function testGetExpressionList()
+    {
+        $compositeExpression = $this->createCompositeExpression();
+
         $expectedExpressionList = array(
             $this->getMock('Doctrine\Common\Collections\Expr\Expression'),
         );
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * @dataProvider invalidDataProvider
+     *
+     * @param $expression
+     * @return void
+     */
+    public function testExceptions($expression)
+    {
+        $this->setExpectedException('\RuntimeException');
```
mikeSimonson: Can you move the setExpêctedException just before the new compositeExpression.
It's just to make sure that it's the new compositeExpression statement that gets tested.

mikeSimonson:
@tobiasoberrauch Thanks a lot

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
```diff
@@ -0,0 +1,112 @@
...
+
+use Doctrine\Common\Collections\Expr\CompositeExpression;
+use Doctrine\Common\Collections\Expr\ExpressionVisitor;
+use Doctrine\Common\Collections\Expr\Value;
+use PHPUnit_Framework_TestCase as TestCase;
+
+/**
+ * Class CompositeExpressionTest
+ *
+ * PHP Version 5
```
stof: this should be removed

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * @category  PHP
+ * @package   Doctrine\Tests\Common\Collections\Expr
```
stof: this project does not use ``@category`` and ``@package`` (duplicating the namespace in ``@package`` brings absolutely no value)

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
```diff
@@ -0,0 +1,112 @@
...
+ * @author    Tobias Oberrauch <hello@tobiasoberrauch.com>
+ *
+ * @covers \Doctrine\Common\Collections\Expr\CompositeExpression
```
stof: This project does not use ``@covers``

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
```diff
@@ -0,0 +1,112 @@
...
+            $expression,
+        );
+
+        $compositeExpression = new CompositeExpression($type, $expressions);
+    }
+
+    /**
+     * @return CompositeExpression
+     */
+    public function testConstructor()
```
stof: This test is not testing anything (there is no assertion), so it should not be a test

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
mikeSimonson: Does it really hurt to be specific about what we are trying to test ?

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
tobiasoberrauch: what do you mean? The use statements or the complete test?

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
tobiasoberrauch: I removed it with the newest commit

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
tobiasoberrauch: Shouldn't we test exceptions?

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
tobiasoberrauch: I changed it

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
stof: This test is not testing anything. It has no assertion.

This is not a test testing exceptions (``testExceptions`` is such a test)

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
mikeSimonson: He mean the PHP version 5

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
mikeSimonson: Please squash your commits and force push them.

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
mikeSimonson: In this test you seems to be testing that the constructor does not throw exception when everything is normal.
To achieve this test your now need to assert at the end of the test that the object returned is indeed of the the type compositeExpression with assertInstanceOf for example.

tobiasoberrauch: Ok, I just removed it

stof: And IMO, we don't need a specific test for that, as other tests will need to instantiate the object anyway, so the working instantiation will already be covered

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
tobiasoberrauch: I got it and I removed the unnecessary comments

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * @author  Tobias Oberrauch <hello@tobiasoberrauch.com>
+ * @covers  \Doctrine\Common\Collections\Expr\CompositeExpression
+ */
+class CompositeExpressionTest extends TestCase
+{
+    /**
+     * @return CompositeExpression
+     */
+    public function testConstructor()
```
stof: I don't understand what you did here. you kept the test which is not testing anything (test without assertion), and you removed the test which was covering the case of exceptions (ensuring we don't accept composite without expressions). What we asked is the opposite

tobiasoberrauch: Now i got it :) I fixed it

tests/Doctrine/Tests/Common/Collections/Expr/CompositeExpressionTest.php
mikeSimonson: uh nope

mikeSimonson:
Thanks

PR #88, opened 2016-06-10
Null as default value of first() and last()

Raphhh:
Change first() and last() methods to return NULL as default value.
Because NULL is the default value of an object when we use type hinting, we can give directly the result of first/last method without worry to know if there is an object in the collection.

So, allow to code this:

```
function foo (Bar $bar = null) { /*...*/ }

foo($collection->first());
```

Instead of:

```
function foo (Bar $bar = null) { /*...*/ }

foo($collection->first() ? $collection->first() : null);
```

### DIFF
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -79,6 +79,9 @@ class ArrayCollection implements Collection, Selectable
      */
     public function last()
     {
+        if (!$this->elements) {
+            return null;
+        }
         return end($this->elements);
     }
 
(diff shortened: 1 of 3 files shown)


Ocramius:
This changes existing semantics. `reset` and `end` return `false`, and existing code will rely on that `false` value, making this PR a subtle and very hard to "comply with" BC break.

Ocramius:
Note that I do not disagree with `null` being returned, it just cannot be applied to the current implementation, since so many rely on it.

Raphhh:
Yes of course it breaks BC. I tought it could be for a new major release. Thank you for your time. :)

flo-sch:
It makes absolute sense for me that an empty collection returns `null` rather than `false`, when calling `$collection->first()` or `$collection->last()` aswell.

Hope this PR will be merged in a major release?

PR #89, opened 2016-06-16
Add Criteria::fromFindCriteria

soullivaneuh:
## Subject

The goals is to be able to quick reproduce a criteria looking like parameters given to ORM find methods.
## Concrete case

On a REST controller action, I would like to setup basic filter and apply them with the `findBy` method of the ORM.

I have a `Server` entity containing many `User` entities.

When the user calling the api method is not an admin, I have to filter the server list to get only related ones.

But the `findBy` does not allow us to do a search on a `ManyToMany` relationship.

I would like to keep same method on criteria to apply it on a query builder:

``` php
public function cgetAction(ParamFetcherInterface $paramFetcher)
{
    $view = $this->view();

    $filters = [
        'name' => $paramFetcher->get('name'),
    ];
    $filters = array_filter($filters, function ($value) {
        return null !== $value;
    });

    if ($this->get('admin.server')->isGranted('LIST')) {
        $servers = $this->getDoctrine()->getRepository('AppBundle:Server')->findBy($filters);
        $view->getSerializationContext()->setGroups(['list', 'list_admin']);
    } else {
        $servers = $this->getDoctrine()->getRepository('AppBundle:Server')->getByOwnerQB($this->getUser())
            ->addCriteria(Criteria::fromFindCriteria($filters))
            ->getQuery()->getResult();
        $view->getSerializationContext()->setGroups('list');
    }

    $view->setData($servers);

    return $view;
}
```

Having this method would simplify the logic.

### DIFF
--- a/lib/Doctrine/Common/Collections/Criteria.php
+++ b/lib/Doctrine/Common/Collections/Criteria.php
@@ -76,6 +76,31 @@ class Criteria
     }
 
     /**
+     * Creates an instance with same behaviour as criteria parameters passed on ORM find methods.
+     *
+     * @param mixed[] $findCriteria Array of different keys and values. Gets the same format as argument expected on
+     *                              Doctrine\Common\Persistence\ObjectRepository::findBy and
+     *                              Doctrine\Common\Persistence\ObjectRepository::findOneBy methods.
+     *
+     * @return Criteria
+     */
+    public static function fromFindCriteria(array $findCriteria)
+    {
+        $criteria = static::create();
+
+        foreach ($findCriteria as $key => $value) {
+            $criteria->andWhere(!is_array($value)
+                ? static::expr()->eq($key, $value)
+                : new CompositeExpression(CompositeExpression::TYPE_OR, \array_map(function ($v) use ($key) {
+                    return static::expr()->eq($key, $v);
+                }, $value))
+            );
+        }
+
+        return $criteria;
+    }
+
+    /**
      * Returns the expression builder.
      *
      * @return \Doctrine\Common\Collections\ExpressionBuilder
--- a/tests/Doctrine/Tests/Common/Collections/CriteriaTest.php
+++ b/tests/Doctrine/Tests/Common/Collections/CriteriaTest.php
@@ -16,6 +16,21 @@ class CriteriaTest extends \PHPUnit_Framework_TestCase
         $this->assertInstanceOf(Criteria::class, $criteria);
     }
 
+    public function testFromFindCriteria() : void
+    {
+        $criteria = Criteria::fromFindCriteria(['name' => 'test', 'foo' => [42, 1337]]);
+
+        /** @var CompositeExpression $where */
+        $where = $criteria->getWhereExpression();
+        $this->assertInstanceOf(CompositeExpression::class, $where);
+
+        $this->assertSame(CompositeExpression::TYPE_AND, $where->getType());
+        $this->assertCount(2, $where->getExpressionList());
+        $this->assertInstanceOf(Comparison::class, $where->getExpressionList()[0]);
+        $this->assertInstanceOf(CompositeExpression::class, $where->getExpressionList()[1]);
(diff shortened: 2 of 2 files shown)


lib/Doctrine/Common/Collections/Criteria.php
```diff
@@ -76,6 +76,24 @@ public static function create()
     }
 
     /**
+     * Creates an instance with same behaviour as criteria parameters passed on ORM find methods.
+     *
+     * @param array $findCriteria
```
greg0ire: "array" => "Criteria[]" ? Not sure what is used in this project…

Soullivaneuh: Nope, this is the same as `ObjectRepository::findBy` method first parameter.

Ref: https://github.com/doctrine/common/blob/faf7d81273eeffb84d524006daf85d37a14ac42a/lib/Doctrine/Common/Persistence/ObjectRepository.php#L64

I'll add a comment.

lib/Doctrine/Common/Collections/Criteria.php
```diff
@@ -76,6 +76,25 @@ public static function create()
     }
 
     /**
+     * Creates an instance with same behaviour as criteria parameters passed on ORM find methods.
+     *
+     * @param mixed[] $findCriteria Array of different keys and values. Gets the same format as argument expected on
+     *                              ObjectRepository::findBy and ObjectRepository::findOneBy methods.
```
stof: you should use fully qualified class names, as they are in a different package

Soullivaneuh:
@stof fixed.

Soullivaneuh:
Any news on this PR?

mikeSimonson:
@Soullivaneuh As far as I can see in [the entityPersister](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L1580), if you want to make the 2 compatible you have to take into account 3 more cases and their tests.

  * [The case were the value passed is an array and convert to IN](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L1621)
  * [The case were the value is an array and one of the value is null](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L1623)
  * [The case were the value is null](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L1635)

Thanks

Soullivaneuh:
@mikeSimonson Okay I'll see to update the tests first.

tests/Doctrine/Tests/Common/Collections/CriteriaTest.php
```diff
@@ -16,6 +16,18 @@ public function testCreate() : void
         $this->assertInstanceOf(Criteria::class, $criteria);
     }
 
+    public function testFromFindCriteria()
+    {
+        $criteria = Criteria::fromFindCriteria(array('name' => 'test', 'foo' => 42));
```
lcobucci: Please use short array syntax

tests/Doctrine/Tests/Common/Collections/CriteriaTest.php
```diff
@@ -16,6 +16,18 @@ public function testCreate() : void
...
+
+        /** @var CompositeExpression $where */
+        $where = $criteria->getWhereExpression();
+        $this->assertInstanceOf('Doctrine\Common\Collections\Expr\CompositeExpression', $where);
```
lcobucci: Please use `::class` syntax

lcobucci (commented):
Please also address this (after a rebase to sync with `master`).

Soullivaneuh:
@mikeSimonson I updated the code to manage the first case.

Stop me if I'm wrong, but I don't thing I have anything to do for the `null` cases.

This will be simply manage by this:

```php
static::expr()->eq($key, $value);
```

Where `$value` will be `null`.

PR #90, opened 2016-06-16
Do not target unbound PHP version contraint

Soullivaneuh:
Ref: https://getcomposer.org/doc/faqs/why-are-unbound-version-constraints-a-bad-idea.md

@Ocramius I heard you hate unbound version constraint on PHP... :stuck_out_tongue:


Ocramius:
@Soullivaneuh thanks!

PR #91, opened 2016-07-08
Added ArrayCollection::createFrom.

Bilge:
This allows derived classes to specify how new instances should be created when constructor semantics have changed. Without this, derived classes that modify constructor semantics have to re-implement the four affected methods using copy and paste because all hard-coded calls to `new static($elements)` are invalid.

:bookmark: I would appreciate a new tag for this PR.


Ocramius:
From the IRC logs, here's the "why" behind this PR:

 * consumer may sub-class `ArrayCollection`
 * sub-class of `ArrayCollection` may define `public function __construct(HahaSomething $else)`
 * these methods would crash due to `new static()` (should use `new self()` or allow delegations, like done in this PRs)

The issue is that `ArrayCollection` has been designed **A LOT OF TIME AGO**, and hasn't been made `final`. This leads to very fragile subclasses, because the design is not extensible by itself.

We can't make `ArrayCollection` `final` though, as it's extensively extended everywhere, and even extended within ORM and ODMs. Therefore we'd break BC after multiple major versions, and with large warning upfront.

The proper solution would be to implement `Collection` in subclasses, but meanwhile, this patch is an acceptable fix. Tests representing the scenario described above are still required here though.

Bilge:
@Ocramius Test added. I also got rid of that shitty bootstrap.

Ocramius:
👍 thanks!

PR #95, opened 2016-11-15
Add same field name method accessor support to ClosureExpressionVisitor#getObjectFieldValue.

obokaman-com:
It's been an extended practice using the same name of a private property for the method returning its value. Actually getObjectFieldValue don't check for the existence of a method with the field name, only is*, get*, the field itself if it's public, or using `__get` or `__call` magic methods.

I've added the implementation and additional tests to include this feature.


Ocramius:
Don't think this sort of approach is valid: getters are getters, and even
then it is risky to call them. Here you are even calling stuff that doesn't
even reflect traditional getter syntax. I, for example, explicitly use this
syntax to break auto-getter-calling systems sometimes.

PR #97, opened 2016-11-28
Enable strict comparison for arrays

vlastv:
I found a bug in the comparison strings starting with 0.


Ocramius:
Possibly a minor BC break, so we'll have to document it on merge/release

Ocramius:
This has been manually merged in doctrine/collections@a01ee38fcd999f34d9bfbcee59dbda5105449cbf

PR #98, opened 2016-12-03
Increase code coverage

snsanich:
Hi.

I applied the same test suite on AbstractLazyCollection as for ArrayCollection (definitely, tests on __toString and Selectable interface was excluded for AbstractLazyCollection)


tests/Doctrine/Tests/LazyArrayCollection.php
```diff
@@ -10,13 +11,20 @@
  */
 class LazyArrayCollection extends AbstractLazyCollection
 {
+    private $_onInitialization;
```
Ocramius: Docblock. Also, please don't prefix private/protected stuff with `_` (we used to do it with PHP 4, when visibility wasn't supported yet)

tests/Doctrine/Tests/Common/Collections/DerivedCollectionTest.php
```diff
@@ -0,0 +1,46 @@
...
+     * class that implements different constructor semantics.
+     */
+    public function testDerivedClassCreation()
+    {
+        $collection = new DerivedArrayCollection(new \stdClass);
+        $closure = function () {
+            // Intentionally empty.
+        };
+
+        self::assertInstanceOf('Doctrine\Tests\DerivedArrayCollection', $collection->map($closure));
```
Ocramius: Please use `::class` instead of string references

tests/Doctrine/Tests/Common/Collections/DerivedCollectionTest.php
```diff
@@ -0,0 +1,46 @@
...
+        self::assertInstanceOf('Doctrine\Tests\DerivedArrayCollection', $collection->filter($closure));
+        self::assertContainsOnlyInstancesOf('Doctrine\Tests\DerivedArrayCollection', $collection->partition($closure));
+        self::assertInstanceOf('Doctrine\Tests\DerivedArrayCollection', $collection->matching(new Criteria));
+    }
+}
```
Ocramius: Empty newline required

tests/Doctrine/Tests/Common/Collections/DerivedCollectionTest.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+namespace Doctrine\Tests\Common\Collections;
+
+use Doctrine\Tests\DerivedArrayCollection;
+
+/**
+ * @author Alexander Golovnya <snsanich@gmail.com>
+ */
+
```
Ocramius: Remove this whitespace

tests/Doctrine/Tests/Common/Collections/DerivedCollectionTest.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /**
+     * Tests that methods that create a new instance can be called in a derived
+     * class that implements different constructor semantics.
+     */
+    public function testDerivedClassCreation()
+    {
+        $collection = new DerivedArrayCollection(new \stdClass);
+        $closure = function () {
+            // Intentionally empty.
```
Ocramius: Could you add the `why` this is intentionally empty?

Ocramius: I assume that this should never actually be called?

tests/Doctrine/Tests/Common/Collections/AbstractLazyArrayCollectionTest.php
```diff
@@ -0,0 +1,45 @@
...
+{
+
+    protected function _buildCollection(array $elements = array())
+    {
+        return new LazyArrayCollection(new ArrayCollection($elements));
+    }
+
+    public function testLazyCollection()
+    {
+        $collection = $this->_buildCollection(array('a', 'b', 'c'));
```
Ocramius: Can you switch to short arrays when you move around these files?

snsanich:
I made necessary changes.
I can miss something, @Ocramius , if I haven't understood you correctly.

Ocramius:
@snsanich looking good, waiting for travis :-)

snsanich:
Php 5.3, 5.4 doesn't support short array syntax and Class name resolution via ::class. I removed php 5.3, 5.4 from Travis.yml

Ocramius:
If we're not testing 5.3 and 5.4, then it should also be gone from `composer.json`. Can you update it there too?

Ocramius:
@snsanich awesome, thanks!

PR #99, opened 2016-12-10
[access] added prev acces method to collections

troublete:
Good afternoon, 
i needed this method in order to make something I wrote work. I thought you might like to implement it in the lib so next time me (or someone who needs this) doesn't need to first retrieve the array and execute the native `prev` method on it.

### DIFF
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -260,4 +260,11 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * @return array
      */
     public function slice($offset, $length = null);
+
+    /**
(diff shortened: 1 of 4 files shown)


lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -260,4 +260,11 @@ public function indexOf($element);
      * @return array
      */
     public function slice($offset, $length = null);
+
+    /**
+     * Moves the internal iterator position to the previous element and returns this element.
+     *
+     * @return mixed
+     */
+    public function prev();
```
Ocramius: This is a major BC break, so we cannot apply t the patch.

troublete: okey dokey

PR #100, opened 2016-12-13
Allow use array and multi arguments

vlastv:


lcobucci:
@vlastv please provide a test case as well. 

BTW why do you need that? Can't you use `...$myArray` on the function call?

vlastv:
Variable-length arguments implemented in PHP 5.6, library can work with PHP 5.5 based on composer.json.

I use in case:
```php
$restrictions = [];
foreach($filters as $field => $value) {
    $restrictions[] = Criteria::expr()->eq($field, $value);
}

$criteria->andWhere(call_user_func_array([Criteria::expr(), 'andX'], $restrictions]));
```

IMHO, it next variant is more readable:
```php
// loop

$criteria->andWhere(Criteria::expr()->andX($restrictions)));
```

Ocramius:
@vlastv can be bumped to PHP 5.6 as minimum requirement

vlastv:
@Ocramius If you change the minimum version then this request can be rejected.

Ocramius:
@vlastv I'm suggesting to add it to this patch ;-)

lib/Doctrine/Common/Collections/ExpressionBuilder.php
```diff
@@ -42,7 +42,7 @@ class ExpressionBuilder
      */
     public function andX($x = null)
     {
-        return new CompositeExpression(CompositeExpression::TYPE_AND, func_get_args());
+        return new CompositeExpression(CompositeExpression::TYPE_AND, is_array($x) ? $x : func_get_args());
```
Ocramius: This should be changed to just accept `$x`, and the method signature should become `...$x`

lib/Doctrine/Common/Collections/ExpressionBuilder.php
```diff
@@ -52,7 +52,7 @@ public function andX($x = null)
      */
     public function orX($x = null)
     {
-        return new CompositeExpression(CompositeExpression::TYPE_OR, func_get_args());
+        return new CompositeExpression(CompositeExpression::TYPE_OR, is_array($x) ? $x : func_get_args());
```
Ocramius: This should be changed to just accept `$x`, and the method signature should become `...$x`

tests/Doctrine/Tests/Common/Collections/ExpressionBuilderTest.php
```diff
@@ -37,6 +37,46 @@ public function testOrX()
...
+            $this->builder->eq("c", "d"),
+        ];
+
+        $expr = $this->builder->andX(
+            $x[0],
+            $x[1]
+        );
+        $this->assertEquals($x, $expr->getExpressionList());
+
+        $expr = $this->builder->andX($x);
```
Ocramius: This seems invalid to me

tests/Doctrine/Tests/Common/Collections/ExpressionBuilderTest.php
```diff
@@ -37,6 +37,46 @@ public function testOrX()
...
+            $this->builder->eq("c", "d"),
+        ];
+
+        $expr = $this->builder->orX(
+            $x[0],
+            $x[1]
+        );
+        $this->assertEquals($x, $expr->getExpressionList());
+
+        $expr = $this->builder->orX($x);
```
Ocramius: This seems invalid to me

Ocramius (changes requested):

lib/Doctrine/Common/Collections/ExpressionBuilder.php
```diff
@@ -40,19 +40,19 @@ class ExpressionBuilder
      *
      * @return CompositeExpression
      */
-    public function andX($x = null)
+    public function andX(...$x)
```
lcobucci: @Ocramius should we use `Expression ...$x` as well (and drop the `mixed`)?

lcobucci: HHVM pffff 😜

lcobucci: @stof are you sure about strings? https://github.com/doctrine/collections/blob/0b0b54166319e02f50f996f8d966aed368a78bfe/lib/Doctrine/Common/Collections/Expr/CompositeExpression.php#L49 seems to say something different...

stof:
The current state of the PR does not change the API at all. It still does not allow to pass a single array instead of variadic arguments. So it has no benefit for people using Doctrine (it does not allow to do what the description says).

@vlastv if your own code runs on PHP 5.6+, you can avoid ``call_user_func_array`` today without any change in Doctrine:

```php
// loop

$criteria->andWhere(Criteria::expr()->andX(...$restrictions)));
```

lib/Doctrine/Common/Collections/ExpressionBuilder.php
vlastv: @stof strings not allowed [here](https://github.com/doctrine/collections/blob/master/lib/Doctrine/Common/Collections/Expr/CompositeExpression.php#L57)

stof: sorry, I confused the Criteria system in doctrine/collection with the expression builder of the QueryBuilder. Forget my comment.

stof:
So the *only* thing achieved by your PR so far is breaking support for HHVM, without any benefit for users of the API...

vlastv:
I close this PR?

lcobucci:
Now that we dropped HHVM support we can rebase this branch and be more strict on the typehint...

Ocramius:
This can only target 2.x, sadly. See https://3v4l.org/QIi9Y - it is a BC break due to the `ExpressionBuilder` being an open class.

PR #101, opened 2016-12-14
[enhancement] Execute matching comparison through external API.

sovaalexandr:
Major breaking change.
Alternative functionality for collection elements comparison.
Provides matching functionality for replacement to OCP-violating `ExpressionVisitor`.


sovaalexandr:
I'd liked to suggest those changes as an alternative to #96 
As @stof mentioned at that [conversation](https://github.com/doctrine/collections/pull/96#issuecomment-266512478) we can't introduce something like `getFilterCallback` - method that will be binded to particular implementation.
So got an idea - make everything in polymorphic way - through exernal `Doctrine\ORM\Query\Expr`-like API.

As you may see at `ClosureMatchClosureTest` it's easily extensible now.

And the main question is - can I hope for merge of such big breaking change at ANY release/milestione and I should send corresponding pull-requests to doctrine ORM/ODM repos or there is no way and I can forget about it?

stof:
Any BC break on the criteria building is a no-go, as it would force to bump doctrine/collections to a new major version, which would only be usable by a new major version of the ORM and ODMs, and there is no such new major version in the work yet.

If you want to have a chance to get this included in a smaller time than several years, build it in a backward compatible way.

sovaalexandr:
Got it.
But first of all - how do you think - how much clients of Doctrine code are hook into it's core/internals? How much of those do implement `Expression` or know about `ExpressionVisitor` that is internal functionality instantiated at it's usage place. I don't think it's a lot - existent collection API lives no sense for this implementations because of switch at `ExpressionVisitor` implementations. Maybe 5% or some.
And second - maybe not directly coupled with current PR, but I have this question for quite a long time and just haven't seen any appropriate moment to ask - you use semantic versioning, as I guess, so what restricts you from releasing new major versions of lib before it's dependents? It shouldn't be problem because everybody fix epoch and major number at composer dependencies. And dependents like ORM/ODM could use new version when those would be ready to.
Pros of such approach:
 * Got bug reports/performance issues before new version will be required by ORM/ODM
 * ORM and ODM could use different collection versions - no need to fit both big repos within a single effort
 * Ease of particular feature usage
Cons:
 * Bit more effort on maintenance
 * If bug occurs at early release - fix should be submitted to every.
 * LTEs and everything connected with it

stof:
@sovaalexandr As the main goal of the doctrine/collection package is to be consumed by the ORM and ODM, releasing a new major version of it years before the ORM and the ODM can move to it does not make sense, as any work benefiting Doctrine should still happen in the old version (and we would have no guarantee that the new version of doctrine/collections would ever meet the needs of the ORM when it is ready to migrate).

And as far as doctrine/collections is concerned, the ORM and ODMs **are** clients dealing with these APIs, and they are released separately so they rely on semver.

> Got bug reports/performance issues before new version will be required by ORM/ODM

that's wrong, as all Doctrine users would still be using the old version as the ORM would forbid them to upgrade to the new version

> ORM and ODM could use different collection versions

This would make the ORM and ODM impossible to use in the same project though, due to a conflicting dependency.

Ocramius:
Closing as per https://github.com/doctrine/collections/pull/101#issuecomment-267387552

PR #103, opened 2016-12-15
Contributing guide

### LINKED ISSUE #102: Contributing guideline
It would be quite nice t have CONTRIBUTING.md at the root of current repository. That would eliminate efforts to find commit check-list, codestyle guides, etc.

sovaalexandr:
[resolves #102 ] Doctrine ORM [CONTRIBUTING.md](https://github.com/doctrine/doctrine2/blob/master/CONTRIBUTING.md) changed to collections context.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,65 @@
+# Contribute to Doctrine
+
+Thank you for contributing to Doctrine!
+
+Before we can merge your Pull-Request here are some guidelines that you need to follow.
+These guidelines exist not to annoy you, but to keep the code base clean,
+unified and future proof.
+
+## We only accept PRs  to "master"
+
+Our branching strategy is "everything to master first", even
+bugfixes and we then merge them into the stable branches. You should only 
+open pull requests against the master branch. Otherwise we cannot accept the PR.
+
+There is one exception to the rule, when we merged a bug into some stable branches
+we do occasionally accept pull requests that merge the same bug fix into earlier
+branches.
+
+## Coding Standard
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+```sh
+git clone git@github.com:doctrine/collections.git
+cd collections
+curl -sS https://getcomposer.org/installer | php --
+./composer.phar install
+```
+
+## Travis
+
+We automatically run your pull request through [Travis CI](http://www.travis-ci.org)
```
stof: I suggest to use ``https://travis-ci.org`` here, to avoid going through a useless redirection.

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+against PHP 5.5, PHP 5.6, PHP 7 and HHVM. If you break the tests, we cannot merge your code,
```
stof: I would say ``against supported PHP versions`` to avoid having to maintain this list when it changes

composer.json
```diff
@@ -13,7 +13,8 @@
         {"name": "Johannes Schmitt", "email": "schmittjoh@gmail.com"}
     ],
     "require": {
-        "php": "^5.5 || ^7.0"
+        "php": "^5.5 || ^7.0",
+        "doctrine/coding-standard": "~0.1@dev"
```
stof: must be a dev requirement. We don't want to install it in each project using the package

sovaalexandr: Thanks a lot. Put it here by mistake - sure it's only dev requirement.

Ocramius:
LGTM 👍

PR #110, opened 2017-02-09
Add hhvm in the allowed failures

mikeSimonson:


Ocramius:
We guarantee HHVM compat. Build is to be fixed instead

mikeSimonson:
@ocramius


As far as I could see here are our options

The main problem is that travis comes with a super old version of hhvm.
We need to update it but it will require sudo.

Then
  * Find the right combination of (hhvm.php7.all=1 and hhvm.php7.builtins=1)
  * Or use `composer install --ignore-platform-reqs` with hhvm but it will maybe have some weird side effect in the dependencies.

Ocramius:
We should try enforcing trusty and a newer hhvm version, as attempted in
other patches

mikeSimonson:
Do you know the version of hhvm on travis in trusty ?
Aka recent enough for hhvm.php7.all=1 and hhvm.php7.builtins=1 ?

Ocramius:
IIRC 3.17

.travis.yml
```diff
@@ -1,18 +1,24 @@
...
 before_script:
-  - composer -n --prefer-source install
+  - if [ $TRAVIS_PHP_VERSION = 'hhvm' ]; then hhvm -d hhvm.php7.builtins=1 composer -n --prefer-source install; else composer -n --prefer-source install; fi
 
 script:
   - vendor/bin/phpunit
 
 after_script:
-  - if [[ $TRAVIS_PHP_VERSION != '7.0' ]] && [[ $TRAVIS_PHP_VERSION != 'hhvm' ]]; then wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover build/logs/clover.xml; fi
+  - if [ $TRAVIS_PHP_VERSION = '7.1' ] && [[ $TRAVIS_PHP_VERSION != 'hhvm' ]]; then wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover build/logs/clover.xml; fi
```
stof: as you use a equality check, there is no need to keep the check for hhvm

update-hhvm-to-php7-compat.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+sudo apt-get install software-properties-common
+sudo apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0x5a16e7281be7a449
+sudo add-apt-repository "deb http://dl.hhvm.com/ubuntu $(lsb_release -sc) main"
+sudo apt-get update
+sudo apt-get install hhvm
```
stof: is this script actually used ?

lcobucci:
We decided to drop HHVM and removed it on #118

PR #111, opened 2017-02-19
Fix DocBlock for derived classes

gadelat:
This goes hand in hand with https://github.com/doctrine/collections/pull/91

If you add new method in your derived class, IDE without this patch doesn't know it's there after you call some fluent method.


Ocramius:
This is an interface change, as it prevents child classes from returning an instance of the parent class.

Closing as `invalid`.

Also note that `static` is not a type hint, while `self` and `parent` are.

PR #112, opened 2017-02-19
Fix DocBlock for derived classes

gadelat:
This goes hand in hand with https://github.com/doctrine/collections/pull/91

If you add new method in your derived class, IDE without this patch doesn't know it's there after you call some fluent method.


Ocramius:
Closing as duplicate: #111

gadelat:
@Ocramius This isn't duplicate. I'm not changing interface now.

Ocramius:
You

Ocramius:
Good to go! Thanks @gadelat

PR #114, opened 2017-03-04
Remove code duplication in remove method using morgan's laws

santakadev:


mikeSimonson:
Thanks

Ocramius:
The reason why code is duplicated here is due to PHP being quite slow at method calls, and not having any inlining functionality in the engine (so far)

Ocramius:
Closing as invalid as per explanation above.

PR #115, opened 2017-03-10
adding reduce capability to collections

rafaelmaza:
as in http://php.net/manual/en/function.array-reduce.php

### DIFF
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -223,6 +223,16 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * @return Collection
      */
     public function map(Closure $func);
+    
+    /**
+     * Applies the given function to each element in the collection and returns
+     * a single value.
+     *
+     * @param Closure $func
+     *
+     * @return Collection
+     */
+    public function reduce(Closure $func);
 
     /**
      * Partitions this collection in two collections according to a predicate.
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -242,6 +242,15 @@ abstract class AbstractLazyCollection implements Collection
     /**
      * {@inheritDoc}
      */
+    public function reduce(Closure $func)
+    {
+        $this->initialize();
+        return $this->collection->reduce($func);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     public function partition(Closure $p)
     {
         $this->initialize();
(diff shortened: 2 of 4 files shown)


lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -223,6 +223,16 @@ public function forAll(Closure $p);
...
+    
+    /**
+     * Applies the given function to each element in the collection and returns
+     * a single value.
+     *
+     * @param Closure $func
+     *
+     * @return Collection
+     */
+    public function reduce(Closure $func);
```
Ocramius: Interface additions are a BC break, sorry.

stof:
@Ocramius should we add a warning about it in the Contributing file ? It may help reducing the number of PRs opened to add new methods

mikeSimonson:
@rafaelmaza Thanks for your contribution. Sadly for technical reason this part of the ORM is really tightly coupled to our user and to the rest of the ORM. We really can't make any change that wouldn't be very Strickland backward compatible.

@stof We should definetly mention that in the contributing file.

stevenwadejr:
Serious question: *why* and *how* is an **addition** of a method a breaking change and not backwards compatible? Changing or removing something, sure, but *adding* a method?

stof:
@stevenwadejr this adds a method in an interface which can be (and *is*)  implemented in other packages than `doctrine/collections`. Adding a new method will mean that any other package now contain a broken class (because it misses the new method)

stevenwadejr:
@stof Thanks. That makes perfect sense. 👍 I appreciate it.

davidstanley01:
@stof Do those other packages only implement the `Collection` interface and not extend either the `ArrayCollection` or `AbstractLazyCollection` classes, too? If so, then I agree - this would be a breaking change. If they also extend those two classes, though, then this is _not_ a BC change as this PR (and #121) both implement the method.

Ocramius:
@davidstanley01 we could make these changes if we had private interfaces, but we don't. As soon as an interface lands in a public package, BC concerns affect it.

stof:
@stevenwadejr see https://symfony.com/bc The Symfony core team did a huge work about documenting exactly about what kind of changes are allowed or no regarding backward compatibility (many people forget about some.
Note that this is related to the BC guarantees we provide in the first section. If you want to turn some `No` into `Yes` there to be even stricter, it requires changing lots of `yes` into `No` about allowed changes.

@davidstanley01 they also implement the interface directly in some cases (which is perfectly valid usage of the package)

davidstanley01:
Appreciate your time and the explanations. Thank you!

If you are open to enhancements for the next version, I can certainly help out. We use Doctrine heavily and adding these convenience methods would help out a bunch. 

Thanks, again.

Ocramius:
@davidstanley01 I think that the collections API will be split out somehow: having the methods on the class itself is likely going to be a problem (this one, in particular) in the future too...

davidstanley01:
@Ocramius Ok. At any rate, I'm happy to help when/if we get to that point. Cheers!

stof:
@Ocramius especially for methods getting a callable as argument: there is no way to move this at DB-level for lazy collections, so have a utility class getting the collection as one of the argument of the method could make sense.

PR #119, opened 2017-05-23
Require PHP 7.1

lcobucci:
Related to https://github.com/doctrine/doctrine2/issues/6425

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: php
 sudo: false
 
 php:
-  - 7.0
   - 7.1
   - nightly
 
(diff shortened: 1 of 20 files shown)


malukenho:
@lcobucci how about `declare(strict_types=1);`?

lcobucci:
@malukenho I like the idea but I rather add it on the major release (along with the type-hints everywhere)...

Ocramius:
Strict types can indeed only be added later in a major version. Checking just the current diff, some upcasts may happen for asc/desc ordering, and that means that it would crash hard on people not strictly passing strings to the API.

Ocramius:
: shipit:

Majkl578:
FYI: See #125 for strict types, etc. 😎

PR #120, opened 2017-06-09
Use of keyword "static::" to call getObjectFieldValue in ClosureExpressionVisitor

kdubuc:
Use of Late static bindings to access ClosureExpressionVisitor::getObjectFieldValue depending of the class for which it is called. Allows ClosureExpressionVisitor.php to to be extended.

### DIFF
--- a/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
+++ b/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
@@ -108,8 +108,8 @@ class ClosureExpressionVisitor extends ExpressionVisitor
         }
 
         return function ($a, $b) use ($name, $next, $orientation) : int {
-            $aValue = ClosureExpressionVisitor::getObjectFieldValue($a, $name);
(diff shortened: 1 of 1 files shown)


Ocramius:
>  Allows ClosureExpressionVisitor.php to to be extended.

The `ClosureExpressionVisitor` is not supposed to be extended.

kdubuc:
Why not ? I want to ```ClosureExpressionVisitor``` can handle my domain objects to perform a ```matching(Criteria $criteria)``` on them. The logic remains the same, just ```getObjectFieldValue```must be adapted. I could implements ```ExpressionVisitor```, but the 90% of the class would be copy / pasted.

Ocramius:
> Why not ?

Because it was not designed for that. The only mistake we did here is not declaring it `final` when we first wrote it.

 > 90% of the class would be copy / pasted.

That's an acceptable trade-off compared to inheritance coupling.

kdubuc:
Okay, I assume the final keyword now 😉 
Thanks for your time (and all OSS projects you lead !)

PR #121, opened 2017-06-14
Add reduce method and tests

davidstanley01:
This pull requests adds a `reduce` method to the `Collection` interface as well as a simple implementation. Additionally, I added a test for this new method.  

Following the pattern set by the `map` method, this is really just a wrapper for the native `array_reduce` function.

Let me know if there is anything you'd like in addition to this.


Ocramius:
Sorry, I have to close this as a duplicate.
https://github.com/doctrine/collections/pull/115

PR #122, opened 2017-06-26
Add missing PHP version bumping to 7.1

Soullivaneuh:
7.1 is required since #119, but composer was not updated.


Ocramius:
LGTM, thanks!

PR #123, opened 2017-07-11
Case insensitive like

okoehler:
This PR adds Comparison::ICONTAINS to create a case insensitive like.

Depends on doctrine [PR#6547](https://github.com/doctrine/doctrine2/pull/6547)


lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -169,6 +169,7 @@ public function walkComparison(Comparison $comparison)
                 };
 
             case Comparison::CONTAINS:
+            case Comparison::CONTAINS_CI:
```
Ocramius: `strpos` is case-sensitive, so this block is invalid

Ocramius (changes requested):

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -173,6 +173,11 @@ public function walkComparison(Comparison $comparison)
                     return false !== strpos(ClosureExpressionVisitor::getObjectFieldValue($object, $field), $value);
                 };
 
+            case Comparison::ICONTAINS:
+                return function ($object) use ($field, $value) {
+                    return false !== strpos(mb_strtolower(ClosureExpressionVisitor::getObjectFieldValue($object, $field)), mb_strtolower($value));
```
Majkl578: Hmm, although it's correct, using mb_strtolower introduces dependency on ext-mbstring, which is not bundled in PHP. It's not hard dependency (required only when ICONTAINS is used), but it should probably be added to _suggest_ section in the composer.json.

okoehler: @Majkl578 Done (PR updated)

lcobucci: Why not using [`mb_stripos()`](http://php.net/manual/en/function.mb-stripos.php)?

isisis:
@Ocramius I'm waiting for this feature as well, what's the current state?

PR #125, opened 2017-07-15
[2.0] Strict types, native parameter & return types

Majkl578:
Migrating code to PHP 7.1:
* opened 2.0.x-dev
* enabled strict types
* added parameter & return types
* added const visibility
* added brief upgrading notes

No changes in tests needed except one inheritance violation.

There is no develop branch yet, thus targetting master - let me know how to proceed or when to change base branch. 😎

### DIFF
--- a/UPGRADING-2.0.md
+++ b/UPGRADING-2.0.md
@@ -1,0 +1,61 @@
+Upgrading from 1.x to 2.0
+=========================
+
+
+## BC breaking changes
+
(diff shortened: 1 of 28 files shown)


psfpro:
Hi! What about the slice method? I think slice should return the Collection like filter method.
This will make collection more useful. For example:
```php
$collection->slice(0, 20)->map($func)->toArray();
```

PR #127, opened 2017-07-24
Use PSR-4

Nyholm:
Sorry for bothering this super stable package. May I ask to consider to update to PSR-4?

Im using https://github.com/humbug/php-scoper which automatically rewrites the namespace before Im building a PHAR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "doctrine/coding-standard": "~0.1@dev"
     },
     "autoload": {
-        "psr-0": { "Doctrine\\Common\\Collections\\": "lib/" }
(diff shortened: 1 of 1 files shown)


Ocramius:
LGTM :+1:

Nyholm:
Thank you merging

greg0ire:
I think this should have been done on 1.5, and the depth of the tree shouldn't have been preserved. See  https://github.com/doctrine/collections/pull/138/files

Ocramius:
Depth of tree is correctly preserved, as the doctrine organisation does not
endorse PSR-4 namespace prefixes and still suggests for the sake of tooling
simplicity to use PSR-0 path mappings.

PR #128, opened 2017-07-31
Make contains comparison case insenstive

mdawart:
Should the comparison be changed to be case insensitive or another case insensitive comparison be added (see https://github.com/doctrine/collections/pull/123)


Ocramius:
Closing as `invalid` here: #123 is the correct patch, and includes tests too.

PR #130, opened 2017-08-29
Update build to add PHPCS and PHPStan

lcobucci:


phpcs.xml.dist
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
+<ruleset name="Coding Standards for doctrine">
+    <description>Coding Standards for doctrine.</description>
```
Majkl578: This should probably be _"Coding Standard for Doctrine Collections."_ (and same for ruleset name)

phpunit.xml.dist
```diff
@@ -1,13 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
-         backupGlobals="false"
-         colors="true"
-         verbose="true"
-         beStrictAboutOutputDuringTests="true"
-         beStrictAboutTestsThatDoNotTestAnything="true"
```
Majkl578: Why was this removed?

lcobucci: I've tried explained on the commit message (cb7f39d) but the reason is that we were just repeating the default configuration, which is kind of redundant.

phpcs.xml.dist
lcobucci: You're right, I was too lazy to change it (and I probably didn't change on the other projects as well - shame on me!).

lcobucci: I've just found out that the `name` attribute and the `description` element are useless here. So I'll just kill them

Issue #134, opened 2017-11-21
ClosureExpressionVisitor not using public method matching field name


rikbruil:
I've run into a situation where I'm using ArrayCollection to filter some objects using Criteria::orderBy().

Assume these objects have a private property 'foo', and a public method 'foo()' which returns the private property value. The ClosureExpressionVisitor will check for various public methods existence (for example is*/get* variants, or access through magic methods like __get/__call).

I have the feeling there is one more check missing here, which is checking if a public method exists that matches the field name literally (without prepending get/is or through magic methods).

I'm not sure if this was omitted with a reason, maybe this means we are using a convention thats not very widespread? I will create a PR to hopefully communicate my issue better.

malarzm:
Relates to https://github.com/doctrine/collections/pull/149

PR #135, opened 2017-11-21
Add check in ClosureExpressionVisitor to use literal public method if available

rikbruil:
The PR belonging to #134.

I'm not sure if there was a reason not to check for existence of a method with the literal passed field name. If not, this PR adds this check ;)


Majkl578:
Internal note: We should consider using naming strategy provided from outside rather than trying to guess method name convention.

albertboada:
👍

jwage:
Seems okay to me.

Ocramius:
Closing as `won't fix` and `bc break` for reasons matching the ones I've explained in #149

PR #138, opened 2018-01-21
Leverage PSR-4 to provide a shallower tree

greg0ire:
This should make fuzzy matchers more efficient.


phpunit.xml.dist
```diff
@@ -8,16 +8,17 @@
          beStrictAboutOutputDuringTests="true"
          beStrictAboutTestsThatDoNotTestAnything="true"
          beStrictAboutTodoAnnotatedTests="true"
+         bootstrap="vendor/autoload.php"
```
greg0ire: I'm sure this is need it but I'm not sure why. Or rather, I'm not sure why it worked before. It's as if phpunit assumed psr 0

greg0ire:
If you are ok with this I might do the same on other Doctrine libs.

lcobucci (changes requested):
Thanks for your effort but the Doctrine team voted against PSR-4 and also decided to not apply the "shallower tree" in any of our projects.

Ocramius:
Same here: shallow tree is not an advantage to me, and the doctrine
organisation voted against PSR-4 specifically due to the introduced tooling
complexity added by path prefixing.

I'd personally still very much like to keep the path 1:1 with the class
names, as it should be.

PSR-4 as an autoloader is fine and already implemented in this package.

On 22 Jan 2018 00:05, "Luís Cobucci" <notifications@github.com> wrote:

*@lcobucci* requested changes on this pull request.

Thanks for your effort but the Doctrine team voted against PSR-4 and also
decided to not apply the "shallower tree" in any of our projects.

In phpunit.xml.dist
<https://github.com/doctrine/collections/pull/138#discussion_r162828848>:

> @@ -8,16 +8,17 @@
          beStrictAboutOutputDuringTests="true"
          beStrictAboutTestsThatDoNotTestAnything="true"
          beStrictAboutTodoAnnotatedTests="true"
+         bootstrap="vendor/autoload.php"

Not needed when PHPUnit is executed from local composer installation

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/doctrine/collections/pull/138#pullrequestreview-90348231>,
or mute the thread

greg0ire:
Closing then

mvorisek:
2021 to rethink this?

greg0ire:
I didn't follow how things evolved, but I noticed that the DBAL does use a shallow tree since 3.0.x so maybe this might be reconsidered indeed.

PR #139, opened 2018-01-28
Apply Doctrine CS 2.1

Majkl578:
BC-breaking sniffs disabled. Also disabled checks for iterable types since Collections are built around mixed values.


Ocramius:
:ship:

PR #149, opened 2018-04-15
can access the field of a object through same name method

oxkhar:
A object in collection with private attribute, and accesor is a method
which name is the same. Closure expression visitor can access value
throught that metod if exists.

### DIFF
--- a/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
+++ b/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
@@ -22,7 +22,7 @@ class ClosureExpressionVisitor extends ExpressionVisitor
 {
     /**
      * Accesses the field of a given object. This field has to be public
-     * directly or indirectly (through an accessor get*, is*, or a magic
+     * directly or indirectly (through an accessor get*, is*, *, or a magic
      * method, __get, __call).
      *
      * @param object|array $object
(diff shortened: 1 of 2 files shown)


alcaeus:
> This is a BC break

I don't think so. The BC break would be preferring `someField()` to `isSomeField()`, which is easily fixed. Besides that, I don't see this as a BC break. Just like `getSomeField` and `isSomeField` are valid accessors for `someField`, so is `someField(?<type> $someField)`. Some people prefer that access pattern for boolean flags as opposed to having `setFlag` and `isFlag` (or `enableFlag`, `disableFlag` and `isFlag`).

Ocramius:
@alcaeus personally speaking, I do use this kind of accessor naming SPECIFICALLY to prevent tooling from auto-discovering it

alcaeus:
> personally speaking, I do use this kind of accessor naming SPECIFICALLY to prevent tooling from auto-discovering it

That sounds like there's a different problem going on, namely these tools being used when they shouldn't. IMO this is a valid access style for properties and shouldn't be used to "hide something from tooling", especially since the Symfony `PropertyAccessor` component already reads it, so this is a very weak protection against unwanted access. Fact is, if you expose a relatively standard accessor (which this is), tooling should be able to use it.

oxkhar:
Personally I made this request because is not strange that your entities define getters accessors at its private properties with a same name method. It's a convention that would be considered in order to get attributes values, and if you prefer I set in last order to be in mind...

Ocramius:
> It's a convention that would be considered in order to get attributes values, and if you prefer I set in last order to be in mind...

I write my accessors specifically as `class Foo { private $bar; function bar() { return $this->bar; }}` to avoid tooling that calls `getBar()` magically.

I'd say that we should deprecate the entire property access via getter instead, and rely only on state for lookups, which would remove the discussion in first place. Still, that is not for me to decide here, but I'm gonna close this PR as `won't fix` for now.

PR #151, opened 2018-05-20
Using array_splice instead of unset

elier:
Using array_splice instead of unset to avoid missing index in the collection that cause issues with json_encode function.

### DIFF
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -123,7 +123,7 @@ class ArrayCollection implements Collection, Selectable
         }
 
         $removed = $this->elements[$key];
-        unset($this->elements[$key]);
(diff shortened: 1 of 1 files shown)


elier:
I was encoding to json an object with his oneToMany relation when I hit this issue:
https://stackoverflow.com/questions/3869129/php-json-encode-as-object-after-php-array-unset

I guess is the json_encode function what should do a better job handling an array with missing index.

malarzm:
Thank you for the PR @elier! Unfortunately this can not be merged as `array_splice` is operating on lists (thus requires `int` as an offset) while collections support associative arrays as well. You need to take care of this in your code for instance using [`getValues`](https://github.com/doctrine/collections/blob/33407e41a31e448426563e33a6edf688461a9a86/lib/Doctrine/Common/Collections/ArrayCollection.php#L249) which internally calls `array_values`.

PR #152, opened 2018-06-05
Enables automatic serialization with json_encode native function

hectorprats:
with this feature, we can now use \json_encode with our Collections.

Before, \json_encode returns null. Now we receive a good response

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -139,6 +139,14 @@ abstract class AbstractLazyCollection implements Collection
     /**
      * {@inheritDoc}
      */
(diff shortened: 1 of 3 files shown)


hectorprats:
done!

hectorprats:
up!, please, review. Is usefull

hectorprats:
How can it open the door to worms if it is native PHP code? It is easier for JMSSerializer to make errors than the native PHP code.

Ocramius:
Closing as `won't fix` here: `Collection` is an in-memory representation, and should most likely stay as such.

Also, there is no guarantee that a `Collection` is serializable, as its API allows for containing a `mixed[]`.

Conversion to another format should be done by any mapper (as suggested above, JMSSerializer and the likes) calling `->toArray()` on it.

PR #153, opened 2018-06-07
Use == for DateTimeInterface *Equals comparisons

TomHAnderson:


TomHAnderson:
phpcs/phpcbf want to change the root change of this PR from == / != to === / !== and this comparator is the reason for this PR.  So while travis is marked as failed this PR is correct as submitted.

alcaeus:
I believe it would be more sensible to ensure both values are `DateTimeInterface` instances and then compare their timestamp using a strict comparison. Using a type-unsafe comparison is no good at this stage.

TomHAnderson:
My project is strongly based around hydrators so I'll use hydrator strategies to properly type my variables but that still doesn't solve this DateTime problem.

Majkl578:
> compare their timestamp using a strict comparison

👍 

Although timestamp is probably not a good idea for generic date time object, it may be out of unix timestamp's range. So comparing i.e. ATOM representation may be better in this case.

TomHAnderson:
Some folks recommend using a DateInterval to compare dates with different time zones, etc.  https://stackoverflow.com/questions/8328416/check-if-two-php-datetime-objects-are-set-to-the-same-date-ignoring-time/8328507

I believe the implementation I've put forward here is on the right track and would like your help in determining the best way to compare dates with possible different timezones.

alcaeus:
> The generated entity does not type cast returned values from getters so an integer field is returned as a string.
>
> [ ... ]
>
> It could be argued that the getter is incorrect and should type cast values based on the metadata. However I do not want to stray into non-generated entities. Now my question is, are the entities the problem or is exact === matching of values the problem?

I mean no offense, but you describe the problem in the last post. I'd go one step further and claim that neither your entities nor the visitor are the problem, but rather your hydrator: your entity defines a field as `int`, your metadata defines your field as `int`, yet your hydrator somehow seems to forget this. Once the hydrator respects metadata, the problem goes away, at least for non-object types.

For object types, there's a different problem, as in "what does equal mean"? Does it mean "their values are equal" or "they are the same object"? Forgoing the "same" approach, when would you consider two objects equal? When they are of the same type? When they share a common interface? When all of their properties (recursively) are equal? The `==` check somehow handles it and tries to do type juggling as good as possible, but the fact of the matter is: once two objects are not of the same type, loose comparison becomes a risky business.

In this case, all we can do is add an exception for certain known types, in this case `DateTimeInterface`: we don't care if it's a mutable object or not, we care about its representation. That's the reason why I suggested the timestamp, because it eliminates one more question: are two `DateTimeInterface` objects really equal when they share the same timestamp but different timezones? This is an implementation detail that shouldn't be handled in this library, but rather in the respective `DateTimeInterface` implementations. Since we can't have nice things, we'll ignore that for a moment and build the exception right here. An example for this would be `Doctrine\Common\Comparable`, which we should also leverage here to allow moving comparisons to the implementation instead of handling everything in here.

Long story short: strict typing is upon us and we shouldn't rely on type-unsafe checks to compare different objects. Instead, we should have predictable behavior, ideally defined in the class being compared, hence my opposition to the approach.

Ocramius:
Gonna close this as `won't fix`: we shouldn't introduce arbitrary rules for arbitrary types, and I've previously rejected similar changes for `doctrine/orm` as well.

Possibly interesting to follow up in #168 instead, which approaches the problem in a more general way (and is still a BC break though)

PR #162, opened 2018-09-21
Add prepend() to Collection interface

alsii:
New function ´prepend()´ has been added to Colelction interface.
Implementations have been added to ArrayCollection and AbstractLazyCollection.
Tests added.


malarzm:
I've tagged this as an Enhancement but to quote @Ocramius 

> The ArrayCollection and Collection API is closed for expansion, because it would be a BC break to do so.

so I wouldn't expect this to be merged in the foreseeable future.

Ocramius:
Indeed, closing as `can't fix`

Issue #165, opened 2018-10-10
Allow ArrayCollection#filter to filter by key, value or both


0x13a:
**Scenario**
I want be able to filter an array by key or both (key, value)
```php
    /// DOESN"T WORK :(
    (new ArrayCollection(['italian' => 'pizza', 'turkish' => 'kebab']))
        ->filter(function($k, $v) {
            return 'turkish' === $k;
        })
    // WORKSS!1!!  
    array_filter(['italian' => 'pizza', 'turkish' => 'kebab'], function($k, $v) {
            return 'turkish' === $k;
        }, ARRAY_FILTER_USE_BOTH);
```

**Suggested Solution**
 [Since PHP 5.6](http://php.net/manual/en/array.constants.php#constant.array-filter-use-key) we are able to specify a flag on `array_filter` which allows to filter the element of the array by passing to the closure that filter the array, key, value or both 

```
array array_filter ( array $array [, callable $callback [, int $flag = 0 ]] )

Flag determining what arguments are sent to callback:

ARRAY_FILTER_USE_KEY - pass key as the only argument to callback instead of the value
ARRAY_FILTER_USE_BOTH - pass both value and key as arguments to callback instead of the value

Default is 0 which will pass value as the only argument to callback instead.
```

**Suggested Implementation**
At the moment ArrayCollection#filter uses the default value 0 which makes the flag feature impossible to use.

```php
-    public function filter(Closure $p)
+    public function filter(Closure $p, int $flag = 0)
    {
-        return $this->createFrom(array_filter($this->elements, $p));
+        return $this->createFrom(array_filter($this->elements, $p, $flag));
    }
```

**Concerns**
I understand that exposing lower level APIs to an higher level of abstraction is not probably a good idea. Yet, the additional feature would be useful overall to the client who uses the method

If you provide some feedback I can submit a PR

0x13a:
@Ocramius Thanks for the reply.

Using the default value for `$flag = 0` does not introduce a BC Break as it is the default behaviour of array_filter. I will update this issue as soon as I have a PR.

Ocramius:
@0x13a I'm suggesting to use the `ARRAY_FILTER_USE_BOTH` as default, whilst keeping the public API of the method unchanged.

0x13a:
That makes a lot of sense

Ocramius:
Handled in #167

PR #167, opened 2018-10-15
Allow `ArrayCollection#filter()` to filter by key and also value

0x13a:
Uses ARRAY_FILTER_USE_BOTH in the underlying array_filter function. Does
not cause BC Break.

Any examples where how to run benchmarks against this?

https://github.com/doctrine/collections/issues/165

### DIFF
--- a/tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
+++ b/tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
@@ -58,6 +58,17 @@ abstract class BaseCollectionTest extends TestCase
         $this->collection->add(1);
         $this->collection->add('foo');
         $this->collection->add(3);
+        $res = $this->collection->filter(static function ($v) {
+            return is_numeric($v);
+        });
+        self::assertEquals([0 => 1, 2 => 3], $res->toArray());
+    }
+
+    public function testFilterByValueAndKey() : void
+    {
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
```diff
@@ -58,10 +58,12 @@ public function testFilter() : void
...
         $this->collection->add(3);
-        $res = $this->collection->filter(static function ($e) {
-            return is_numeric($e);
+        $this->collection->add(4);
+        $this->collection->add(5);
+        $res = $this->collection->filter(static function ($v, $k) {
+            return is_numeric($v) && $k % 2 === 0;
         });
-        self::assertEquals([0 => 1, 2 => 3], $res->toArray());
+        self::assertEquals([0 => 1, 2 => 3, 4 => 5], $res->toArray());
```
Ocramius: Please do add a test: don't modify existing tests

0x13a: oops

Ocramius:
Also to be checked: preservation of order, preservation of key and value types (use `self::assertSame()` rather than `self::assertEquals()`)

Ocramius:
Thanks @0x13a!

0x13a:
cheers

PR #174, opened 2018-12-12
Late Static Binding for the ClosureExpressionVisitor

martixy:
Makes the class easier to inherit from.

### DIFF
--- a/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
+++ b/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
@@ -101,8 +101,8 @@ class ClosureExpressionVisitor extends ExpressionVisitor
         }
 
         return static function ($a, $b) use ($name, $next, $orientation) : int {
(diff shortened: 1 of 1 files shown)


Ocramius:
This class is not designed for inheritance: closing as `invalid`.

Ocramius:
> Why not?

BC (backwards compatibility) and maintenance boundaries, mostly. The class you want to extend is `ExpressionVisitor`, not `ClosureExpressionVisitor`.

The mistake is on our end, for not marking `ClosureExpressionVisitor` `final` in first place (and we can't do that now), sorry.

martixy:
The functionality is generic enough to be usable outside doctrine, and reimplementing the whole class seems pointless.

But since I can't do anything I guess I'll settle for copy-pasting it for my purposes.

Ocramius:
> But since I can't do anything I guess I'll settle for copy-pasting it for my purposes.

Please remember to also copy the related unit tests to prevent regressions due to inheritance changes.

PR #176, opened 2019-03-01
Preserve keys in ArrayCollection for Criteria firstresults/maxresults

SenseException:
When a `Crieria` is used with a set firstresult or maxresult, the keys of the resulting ArrayCollection are not preserved.

Because keys are kept on `map` or `filter` too, I expect `matching`, like with the order criteria, to keep the keys with a criteria that slices a result.

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -301,17 +301,17 @@ matching
 --------
 
 Selects all elements from a selectable that match the expression and
-returns a new collection containing these elements.
+returns a new collection containing these elements and preserved keys.
 
 .. code-block:: php
     use Doctrine\Common\Collections\Criteria;
     use Doctrine\Common\Collections\Expr\Comparison;
 
     $collection = new ArrayCollection([
-        [
+        'wage' => [
             'name' => 'jwage',
         ],
-        [
+        'roman' => [
             'name' => 'romanb',
         ],
     ]);
@@ -322,6 +322,6 @@ returns a new collection containing these elements.
 
     $criteria->where($expr);
 
-    $matched = $collection->matching($criteria); // ['jwage']
+    $matchingCollection = $collection->matching($criteria); // [ 'wage' => [ 'name' => 'jwage' ]]
 
 You can read more about expressions :ref:`here <expressions>`.
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -403,7 +403,7 @@ class ArrayCollection implements Collection, Selectable
         $length = $criteria->getMaxResults();
 
         if ($offset || $length) {
-            $filtered = array_slice($filtered, (int) $offset, $length);
+            $filtered = array_slice($filtered, (int) $offset, $length, true);
         }
 
         return $this->createFrom($filtered);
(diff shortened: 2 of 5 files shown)


SenseException:
I have to take a look at tests to make them work. PR is not ready for review.

SenseException:
Yes, I agree with the BC break. `array_slice` preserves string keys even without the fourth argument, which is inconsistent to the numeric keys situation. After my changes, numeric and string keys behave the same way.

alcaeus:
I'm not sure if this is in fact expected behaviour: when slicing an array I expect a new array as the result, which by definition should start with `0` as the first index. Thus, `matching` should reindex items for arrays (e.g. only sequential indexes starting with `0`) and preserve keys for everything else. Not sure if it's feasible to check whether indexes are sequential or if it should only check whether the first item has an an int key with value `0`.

SenseException:
Slicing is an implementation choice in `matching`. Preserving keys goes back to commits in 2014 (https://github.com/doctrine/collections/commit/37eee8be8283e94915d37b6d64fab1c1af869736), where a change in the sorting-behaviour was made to preserve keys.

As a user, I would expect `matching` to preserve keys, because even though it may reduce the collection and return a new `ArrayCollection`, I would expect the same elements (keys and values) in it.

alcaeus:
I'm absolutely with you when it comes to maps, but if I slice an array I expect another array in return. The problem is that we don't really have an array implementation of collection, only a hybrid array/map. So I guess it makes sense to keep behaviour consistent across various methods.

SenseException:
What I want to achieve is a consistent behaviour. Of course we can talk about dropping the original keys completely for `matching()` results, but I guess it would be the bigger BC break between this and preserving keys, considering the past years of experience while I helped people with Doctrine Collections. My decision to preserve the keys came from the fact that

* there was at least one accepted PR to preserve keys (link above)
* Doctrine ORM Query preserves keys using `setFirstResult()` and `setMaxResults()` when the DQL `INDEX BY` is used

Another library can have a different behaviour since it is a different project, but I've decided it would make it easier for a developer when similar methods act in a similar way. A little bit of DX, if you want to call it that way.

SenseException:
I did a rebase to the current master. Should I add the documentation in this PR?

Ocramius:
@SenseException docs maybe to be added to the docblock only? Also, we'll need to decide when to call `1.x` frozen.

SenseException:
I would prefer the information in the documentation, but adding them in addition to the docblock is okay. Reading code should be a last resort for a user.

Ocramius:
> Reading code should be a last resort for a user.

It's generally the first thing I do, before docs or anything else...

SenseException:
Me too, but that's not how everyone roles. And they shouldn't rely on implementation-details to use a library.

Ocramius:
A docblock is not an implementation detail: it's a declaration/part of the interface (in PHP, sadly).

Implementation is what you'd find *inside* a method or constant.

SenseException:
I didn't meant the docblock, I was reffering to "reading code". Reading code should be a last ressort. I want to add an entry in the documentation and additionally in the docblocks.

SenseException:
I've added the docblock comment and the info in the documentation.

PR #177, opened 2019-03-13
Generify collections using Psalm

nschoellhorn:
I've talked to @Ocramius and we find that generics/template types would make a very good addition to this library. Since PHP doesn't come with that feature natively, I have decided to implement it using Psalm and [Psalm's template annotations](https://psalm.dev/docs/templated_annotations/).  

**How is that useful?**
A lot of people already use static analysis to reduce the risk of introducing bugs in your application. Psalm is one of the most popular tools out there to do that. By adding the annotations to Psalm and using the correct `@psalm-var` annotations when using the collections in projects, Psalm makes sure that you only put values of the right type in your collection automatically, since it also scans the `vendor` directory.  
I have created a very minimal proof of concept project over at [nschoellhorn/psalm-collections-test](https://github.com/nschoellhorn/psalm-collections-test). You can clone that, install the dependencies and run `./vendor/bin/psalm` to check it out.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,9 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"},
         {"name": "Johannes Schmitt", "email": "schmittjoh@gmail.com"}
     ],
+    "suggest": {
+        "vimeo/psalm": "Psalm allows you to enforce template types for your collections."
+    },
     "require": {
         "php": "^7.1.3"
     },
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         {"name": "Johannes Schmitt", "email": "schmittjoh@gmail.com"}
     ],
     "require": {
-        "php": "^7.1"
+        "php": "^7.1.3"
     },
     "require-dev": {
         "phpunit/phpunit": "^7.0",
--- a/composer.json
+++ b/composer.json
@@ -17,9 +17,6 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"},
         {"name": "Johannes Schmitt", "email": "schmittjoh@gmail.com"}
     ],
-    "suggest": {
-        "vimeo/psalm": "Psalm allows you to enforce template types for your collections."
-    },
     "require": {
         "php": "^7.1.3"
     },
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "phpunit/phpunit": "^7.0",
         "doctrine/coding-standard": "^6.0",
         "phpstan/phpstan-shim": "^0.9.2",
-        "vimeo/psalm": "^3.1"
+        "vimeo/psalm": "^3.2"
     },
     "autoload": {
         "psr-4": { "Doctrine\\Common\\Collections\\": "lib/Doctrine/Common/Collections" }
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "phpunit/phpunit": "^7.0",
         "doctrine/coding-standard": "^6.0",
         "phpstan/phpstan-shim": "^0.9.2",
-        "vimeo/psalm": "^3.2"
+        "vimeo/psalm": "^3.2.2"
     },
     "autoload": {
         "psr-4": { "Doctrine\\Common\\Collections\\": "lib/Doctrine/Common/Collections" }
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
     "require-dev": {
         "phpunit/phpunit": "^7.0",
         "doctrine/coding-standard": "^6.0",
-        "phpstan/phpstan-shim": "^0.9.2"
+        "phpstan/phpstan-shim": "^0.9.2",
+        "vimeo/psalm": "^3.1"
     },
     "autoload": {
         "psr-4": { "Doctrine\\Common\\Collections\\": "lib/Doctrine/Common/Collections" }
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -8,6 +8,7 @@ use Closure;
  * Lazy collection that is backed by a concrete collection
  *
  * @template T
+ * @template-extends Collection<T>
  */
 abstract class AbstractLazyCollection implements Collection
 {
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -8,7 +8,7 @@ use Closure;
  * Lazy collection that is backed by a concrete collection
  *
  * @template T
- * @template-extends Collection<T>
+ * @template-implements Collection<T>
  */
 abstract class AbstractLazyCollection implements Collection
 {
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -7,8 +7,8 @@ use Closure;
 /**
  * Lazy collection that is backed by a concrete collection
  *
+ * @template TKey of array-key
  * @template T
- * @template TKey
  * @template-implements Collection<TKey,T>
  */
 abstract class AbstractLazyCollection implements Collection
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -6,12 +6,15 @@ use Closure;
 
 /**
  * Lazy collection that is backed by a concrete collection
+ *
+ * @template T
  */
 abstract class AbstractLazyCollection implements Collection
 {
     /**
      * The backed collection to use
      *
+     * @psalm-var Collection<T>
      * @var Collection
      */
     protected $collection;
@@ -327,6 +330,8 @@ abstract class AbstractLazyCollection implements Collection
 
     /**
      * Initialize the collection
+     *
+     * @return void
      */
     protected function initialize()
     {
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -8,14 +8,15 @@ use Closure;
  * Lazy collection that is backed by a concrete collection
  *
  * @template T
- * @template-implements Collection<T>
+ * @template TKey
+ * @template-implements Collection<TKey,T>
  */
 abstract class AbstractLazyCollection implements Collection
 {
     /**
      * The backed collection to use
      *
-     * @psalm-var Collection<T>
+     * @psalm-var Collection<TKey,T>
      * @var Collection
      */
     protected $collection;
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -31,12 +31,15 @@ use function uasort;
  * and may break when we change the internals in the future. If you need to
  * serialize a collection use {@link toArray()} and reconstruct the collection
  * manually.
+ *
+ * @template T
  */
 class ArrayCollection implements Collection, Selectable
 {
     /**
      * An array containing the entries of this collection.
      *
+     * @psalm-var T[]
      * @var array
      */
     private $elements;
@@ -45,6 +48,8 @@ class ArrayCollection implements Collection, Selectable
      * Initializes a new ArrayCollection.
      *
      * @param array $elements
+     *
+     * @psalm-param T[]
      */
     public function __construct(array $elements = [])
     {
@@ -60,6 +65,9 @@ class ArrayCollection implements Collection, Selectable
      * @param array $elements Elements.
      *
      * @return static
+     *
+     * @psalm-param T[]
+     * @psalm-return ArrayCollection<T>
      */
     protected function createFrom(array $elements)
     {
@@ -301,6 +309,9 @@ class ArrayCollection implements Collection, Selectable
      * {@inheritDoc}
      *
      * @return static
+     *
+     * @template U
+     * @psalm-return ArrayCollection<U>
      */
     public function map(Closure $func)
     {
@@ -311,6 +322,9 @@ class ArrayCollection implements Collection, Selectable
      * {@inheritDoc}
      *
      * @return static
+     *
+     * @template U
+     * @psalm-return ArrayCollection<U>
      */
     public function filter(Closure $p)
     {
@@ -397,7 +411,9 @@ class ArrayCollection implements Collection, Selectable
                 $next = ClosureExpressionVisitor::sortByField($field, $ordering === Criteria::DESC ? -1 : 1, $next);
             }
 
-            uasort($filtered, $next);
+            if ($next) {
+                uasort($filtered, $next);
+            }
         }
 
         $offset = $criteria->getFirstResult();
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -33,6 +33,8 @@ use function uasort;
  * manually.
  *
  * @template T
+ * @template-implements Collection<T>
+ * @template-implements Selectable<T>
  */
 class ArrayCollection implements Collection, Selectable
 {
@@ -49,11 +51,31 @@ class ArrayCollection implements Collection, Selectable
      *
      * @param array $elements
      *
-     * @psalm-param T[]
+     * @psalm-param array<T> $elements
      */
     public function __construct(array $elements = [])
     {
         $this->elements = $elements;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @psalm-return array<T>
+     */
+    public function toArray()
+    {
+        return $this->elements;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @psalm-return T
+     */
+    public function first()
+    {
+        return reset($this->elements);
     }
 
     /**
@@ -66,7 +88,7 @@ class ArrayCollection implements Collection, Selectable
      *
      * @return static
      *
-     * @psalm-param T[]
+     * @psalm-param array<T> $elements
      * @psalm-return ArrayCollection<T>
      */
     protected function createFrom(array $elements)
@@ -76,22 +98,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
-     */
-    public function toArray()
-    {
-        return $this->elements;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    public function first()
-    {
-        return reset($this->elements);
-    }
-
-    /**
-     * {@inheritDoc}
+     *
+     * @psalm-return T
      */
     public function last()
     {
@@ -108,6 +116,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-return T
      */
     public function next()
     {
@@ -116,6 +126,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-return T
      */
     public function current()
     {
@@ -124,6 +136,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-return T|null
      */
     public function remove($key)
     {
@@ -139,6 +153,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-param T $element
      */
     public function removeElement($element)
     {
@@ -167,6 +183,8 @@ class ArrayCollection implements Collection, Selectable
      * Required by interface ArrayAccess.
      *
      * {@inheritDoc}
+     *
+     * @psalm-return T|null
      */
     public function offsetGet($offset)
     {
@@ -177,6 +195,8 @@ class ArrayCollection implements Collection, Selectable
      * Required by interface ArrayAccess.
      *
      * {@inheritDoc}
+     *
+     * @psalm-param T $value
      */
     public function offsetSet($offset, $value)
     {
@@ -209,6 +229,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-param T $element
      */
     public function contains($element)
     {
@@ -217,6 +239,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-param Closure(int|string, T):bool $p
      */
     public function exists(Closure $p)
     {
@@ -231,6 +255,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-param T $element
      */
     public function indexOf($element)
     {
@@ -239,6 +265,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-return T|null
      */
     public function get($key)
     {
@@ -255,6 +283,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-return array<T>
      */
     public function getValues()
     {
@@ -271,6 +301,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-param T $value
      */
     public function set($key, $value)
     {
@@ -279,6 +311,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-param T $element
      */
     public function add($element)
     {
@@ -299,6 +333,8 @@ class ArrayCollection implements Collection, Selectable
      * Required by interface IteratorAggregate.
      *
      * {@inheritDoc}
+     *
+     * @psalm-return ArrayIterator<int|string, T>
      */
     public function getIterator()
     {
@@ -311,7 +347,8 @@ class ArrayCollection implements Collection, Selectable
      * @return static
      *
      * @template U
-     * @psalm-return ArrayCollection<U>
+     * @psalm-param Closure(T):U $func
+     * @psalm-return Collection<U>
      */
     public function map(Closure $func)
     {
@@ -323,8 +360,8 @@ class ArrayCollection implements Collection, Selectable
      *
      * @return static
      *
-     * @template U
-     * @psalm-return ArrayCollection<U>
+     * @psalm-param Closure(T):bool $p
+     * @psalm-return Collection<T>
      */
     public function filter(Closure $p)
     {
@@ -333,6 +370,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-param Closure(int|string, T):bool $p
      */
     public function forAll(Closure $p)
     {
@@ -347,6 +386,9 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-param Closure(int|string, T):bool $p
+     * @psalm-return Collection<T>[]
      */
     public function partition(Closure $p)
     {
@@ -383,6 +425,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-return array<T>
      */
     public function slice($offset, $length = null)
     {
@@ -391,6 +435,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-return Collection<T>
      */
     public function matching(Criteria $criteria)
     {
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -410,9 +410,7 @@ class ArrayCollection implements Collection, Selectable
                 $next = ClosureExpressionVisitor::sortByField($field, $ordering === Criteria::DESC ? -1 : 1, $next);
             }
 
-            if ($next) {
-                uasort($filtered, $next);
-            }
+            uasort($filtered, $next);
         }
 
         $offset = $criteria->getFirstResult();
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -32,6 +32,7 @@ use function uasort;
  * serialize a collection use {@link toArray()} and reconstruct the collection
  * manually.
  *
+ * @template TKey of string|int
  * @template T
  * @template-implements Collection<T>
  * @template-implements Selectable<T>
@@ -60,8 +61,6 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
-     *
-     * @psalm-return array<T>
      */
     public function toArray()
     {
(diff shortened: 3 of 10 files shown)


Ocramius:
Huge :+1: from me on this.

This is meta-code that provides immediate value for downstream consumers that use Psalm, without any runtime being affected. The only possible issues that might arise are with `doctrine/annotations` and people that are (erroneously) trying to parse them from `vendor/`

Ocramius:
To be checked: scrutinizer-ci failure (can't run `composer install`)

nschoellhorn:
Needed to bump the PHP version to allow the installation of Psalm to complete. Should be no issue though, since it went from 7.1.0 only up to 7.1.3

Ocramius:
@muglug @weirdan can you glance over this, if you have some time?

weirdan:
@Ocramius sure, will do.

lib/Doctrine/Common/Collections/AbstractLazyCollection.php
```diff
@@ -6,12 +6,16 @@
 
 /**
  * Lazy collection that is backed by a concrete collection
+ *
+ * @template T
+ * @template-extends Collection<T>
```
weirdan:
```suggestion
 * @template-implements Collection<T>
```

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -23,6 +23,10 @@
  * You can not rely on the internal iterator of the collection being at a certain
  * position unless you explicitly positioned it before. Prefer iteration with
  * external iterators.
+ *
+ * @template T
+ * @template-extends IteratorAggregate<int|string, T>
+ * @template-extends ArrayAccess<int|string, T>
```
weirdan:
```suggestion
 * @template-extends ArrayAccess<int|string|null, T>
```
Adding null to the list of key types will allow to push elements into colleciton with empty bracket syntax:
```php
$coll[] = $elt;
```

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -23,6 +23,10 @@
  * You can not rely on the internal iterator of the collection being at a certain
  * position unless you explicitly positioned it before. Prefer iteration with
  * external iterators.
+ *
+ * @template T
+ * @template-extends IteratorAggregate<int|string, T>
```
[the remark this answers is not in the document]
muglug: Yeah, I like `array-key`, but not particularly bothered

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -32,6 +36,8 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * @param mixed $element The element to add.
      *
      * @return bool Always TRUE.
+     *
+     * @psalm-param T $element
```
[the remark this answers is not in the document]
muglug: Can implementers of this function have it return `false`?

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -109,6 +123,8 @@ public function getKeys();
      *
      * @return array The values of all elements in the collection, in the order they
      *               appear in the collection.
+     *
+     * @psalm-return array<T>
```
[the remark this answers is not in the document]
muglug: Psalm doesn't have issues with missing generic types (at least not as far as I'm aware). It always adds in `array-key` when there's only one generic given for `array`

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -170,6 +198,8 @@ public function next();
      * @param Closure $p The predicate.
      *
      * @return bool TRUE if the predicate is TRUE for at least one element, FALSE otherwise.
+     *
+     * @psalm-param Closure(int|string, T):bool $p
```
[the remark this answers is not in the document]
muglug: `@psalm-param Closure(int|string=, T=):bool $p` should have the same effect I think?

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -197,6 +232,10 @@ public function forAll(Closure $p);
      * a new collection with the elements returned by the function.
      *
      * @return Collection
+     *
+     * @template U
+     * @psalm-param Closure(T):U $func
```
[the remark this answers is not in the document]
muglug: `@psalm-param Closure(T=):U $func` should also work here

composer.json
```diff
@@ -18,12 +18,13 @@
...
     "require": {
-        "php": "^7.1"
+        "php": "^7.1.3"
     },
     "require-dev": {
         "phpunit/phpunit": "^7.0",
         "doctrine/coding-standard": "^5.0",
-        "phpstan/phpstan-shim": "^0.9.2"
+        "phpstan/phpstan-shim": "^0.9.2",
+        "vimeo/psalm": "^3.1"
```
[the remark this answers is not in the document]
nschoellhorn: That's an awesome idea, will add that for sure!

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -209,6 +248,9 @@ public function map(Closure $func);
      * @return Collection[] An array with two elements. The first element contains the collection
      *                      of elements where the predicate returned TRUE, the second element
      *                      contains the collection of elements where the predicate returned FALSE.
+     *
+     * @psalm-param Closure(int|string, T):bool $p
+     * @psalm-return Collection<T>[]
```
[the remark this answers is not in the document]
muglug: You can also use `@psalm-return array{0: Collection<T>, 1: Collection<T>}` for explicitness

lib/Doctrine/Common/Collections/Criteria.php
```diff
@@ -61,6 +61,8 @@ public static function expr()
      * @param string[]|null $orderings
      * @param int|null      $firstResult
      * @param int|null      $maxResults
+     *
+     * @return void
```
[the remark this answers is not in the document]
muglug: Yup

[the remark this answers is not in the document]
nschoellhorn: That's not added for Psalm but in general, since no return type was in there before. Just thought I'dd add that too since I saw it while skimming the file. Can remove it though, if you don't want it in there! :)

lib/Doctrine/Common/Collections/Collection.php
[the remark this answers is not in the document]
weirdan: Unfortunately Psalm can't properly analyze union of closures yet (see vimeo/psalm#1462), so you can ignore this (and other suggestions involving such unions).

nschoellhorn:
Thank you for the review guys, especially @weirdan for all the suggestions. Going through them atm, but I think we should agree on whether to use `array-key` or an additional template parameter for the collection key? I am currently more in favor of `array-key` since it comes closer to Java's Collection API, which I am generally a fan of. But that's very subjective. What do you guys think?

lib/Doctrine/Common/Collections/Collection.php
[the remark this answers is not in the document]
nschoellhorn: Better safe than sorry, will add that.

Ocramius:
> or an additional template parameter for the collection key?

The key can be `int|string`, and should be a template parameter, because the possible values for it are `int`, `string` and `int|string`, as far as I know.

lib/Doctrine/Common/Collections/Collection.php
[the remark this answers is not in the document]
weirdan: Judging by the docblock part above they can't (`@return bool Always TRUE.`). But if they can, then this suggestion should be ignored.

lib/Doctrine/Common/Collections/Collection.php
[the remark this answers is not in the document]
weirdan: @muglug maybe. Does `=` signify optional arguments?

lib/Doctrine/Common/Collections/Criteria.php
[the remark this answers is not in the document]
weirdan: I don't have a strong opinion here. Generally, constructor return types look weird to me, even though it's possible to return values from a constructor in PHP.

lib/Doctrine/Common/Collections/Collection.php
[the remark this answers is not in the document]
nschoellhorn: Awesome, didn't know that can be done. Will add this!

weirdan:
@muglug if all those `ArrayCollection` docblocks were replaced with `{@inheritDoc}`, would Psalm inherit the signatures from `Collection` interface?

lib/Doctrine/Common/Collections/Criteria.php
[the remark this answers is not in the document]
nschoellhorn: Oh wow, sorry. Didn't see we're talking about the constructor here. I agree that this shouldn't be added for a constructor. Removing it.

weirdan:
@nschoellhorn You will need to adjust ArrayCollection signatures as well (things like `first`, `next`, etc).

ondrejmirtes:
>  could come up with an annotation that doesn't include a tool prefix

Yeah, I'm usually more aggressive about annotation adoption, I'm not worried about using new syntax in at-param, at-var etc. Thanks to shoving intersection types there, they are currently supported by PhpStorm and other tools, which was my goal (driving innovation).

nschoellhorn:
> @muglug if all those `ArrayCollection` docblocks were replaced with `{@inheritDoc}`, would Psalm inherit the signatures from `Collection` interface?

Sadly, Psalm does not seem to catch that. I tried that with my test repository that I linked above and when running Psalm after removing the annotations from `ArrayCollection::add()`, it doesn't complain anymore. So I think we have to keep the annotations in there.

> @nschoellhorn You will need to adjust ArrayCollection signatures as well (things like `first`, `next`, etc).

Yeah, still at it. Going for an additional template parameter `TKey as array-key` now, so people can either make a collection of type `Collection<array-key, something>` if they don't care about the exact type of their key or they can restrict to either `int` or `string` if they like.

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -31,12 +31,17 @@
  * and may break when we change the internals in the future. If you need to
  * serialize a collection use {@link toArray()} and reconstruct the collection
  * manually.
+ *
+ * @template T
```
alcaeus: For the record, I'd prefer if the template would allow specifying a key type as well as the value type: while in many cases you are using the collection as a list with numeric indexes, MongoDB ODM effectively allows using collections as a hash with string indexes. Instead of using `array-key` or `int|string` generically for keys, users should have to consciously specify whether a collection is indexed with numeric or string keys.

ostrolucky:
> Sadly, Psalm does not seem to catch that. So I think we have to keep the annotations in there.

While I'm all in for adding psalm specific notations here (since there doesn't exist any standard for this so far), I'm not in favor of incorporating psalm bug specific workarounds. This is something that should be fixed by psalm, not us having to copy these to every subclass.

Ocramius:
> I'm not in favor of incorporating psalm bug specific workarounds. 

Must back this too: `doctrine/collections`, being very central to the PHP ecosystem, should not have workarounds, as the types do indeed propagate statically, which will affect downstream consumers. Therefore, while this patch can indeed be polished and finished up, we'll need to address the upstream issue in psalm before merging.

nschoellhorn:
I'd also prefer that. Maybe @muglug can check if there's something that I've done wrong, causing the inheritance not to work properly. If it's not my fault, I'd offer submitting an issue to Psalm.

Ocramius:
Well, that seems to be fixed: https://github.com/vimeo/psalm/releases/tag/3.2

muglug:
Yup, thanks for letting me know! Templated inheritance is non-trivial. I already had it working for templated return types, but had neglected templated param types.

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -109,6 +131,8 @@ public function getKeys();
      *
      * @return array The values of all elements in the collection, in the order they
      *               appear in the collection.
+     *
+     * @psalm-return array<array-key,T>
```
weirdan:
```suggestion
     * @psalm-return T[]
```

weirdan: To be consistent with `getKeys()`

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -170,6 +209,8 @@ public function next();
      * @param Closure $p The predicate.
      *
      * @return bool TRUE if the predicate is TRUE for at least one element, FALSE otherwise.
+     *
+     * @psalm-param Closure(TKey, T):bool $p
```
weirdan:
```suggestion
     * @psalm-param Closure(TKey=, T=):bool $p
```

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -180,6 +210,9 @@ public function exists(Closure $p);
      * @param Closure $p The predicate used for filtering.
      *
      * @return Collection A collection with the results of the filter operation.
+     *
+     * @psalm-param Closure(T):bool $p
```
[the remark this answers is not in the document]
weirdan:
```suggestion
     * @psalm-param Closure(T=):bool $p
```

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -189,6 +233,8 @@ public function filter(Closure $p);
      * @param Closure $p The predicate.
      *
      * @return bool TRUE, if the predicate yields TRUE for all elements, FALSE otherwise.
+     *
+     * @psalm-param Closure(TKey, T):bool $p
```
weirdan:
```suggestion
     * @psalm-param Closure(TKey=, T=):bool $p
```

lib/Doctrine/Common/Collections/Collection.php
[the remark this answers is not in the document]
weirdan:
```suggestion
     * @psalm-param Closure(T=):U $func
```

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -197,6 +243,10 @@ public function forAll(Closure $p);
      * a new collection with the elements returned by the function.
      *
      * @return Collection
+     *
+     * @template U
+     * @psalm-param Closure(T):U $func
+     * @psalm-return Collection<array-key, U>
```
weirdan:
```suggestion
     * @psalm-return Collection<TKey, U>
```
Map is expected to keep keys.

lib/Doctrine/Common/Collections/Selectable.php
```diff
@@ -13,6 +13,8 @@
  * utilizing the query APIs, for example SQL in the ORM. Applications using
  * this API can implement efficient database access without having to ask the
  * EntityManager or Repositories.
+ *
+ * @template T
```
weirdan:
```suggestion
 * @template T
 * @template TKey
```

lib/Doctrine/Common/Collections/Selectable.php
```diff
@@ -21,6 +23,8 @@ interface Selectable
      * returns a new collection containing these elements.
      *
      * @return Collection
+     *
+     * @psalm-return Collection<T>
```
weirdan:
```suggestion
     * @psalm-return Collection<TKey,T>
```

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -310,6 +321,8 @@ public function map(Closure $func)
      * {@inheritDoc}
      *
      * @return static
+     *
+     * @psalm-return ArrayCollection<T>
```
weirdan:
```suggestion
     * @psalm-return ArrayCollection<TKey,T>
```

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -82,6 +75,24 @@ public function first()
...
+     *
+     * This method is provided for derived classes to specify how a new
+     * instance should be created when constructor semantics have changed.
+     *
+     * @param array $elements Elements.
+     *
+     * @return static
+     *
+     * @psalm-param array<T> $elements
+     * @psalm-return ArrayCollection<T>
```
weirdan:
```suggestion
     * @psalm-return ArrayCollection<TKey,T>
```

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -82,6 +75,24 @@ public function first()
...
+     * Creates a new instance from the specified elements.
+     *
+     * This method is provided for derived classes to specify how a new
+     * instance should be created when constructor semantics have changed.
+     *
+     * @param array $elements Elements.
+     *
+     * @return static
+     *
+     * @psalm-param array<T> $elements
```
weirdan:
```suggestion
     * @psalm-param array<TKey,T> $elements
```

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -45,27 +51,14 @@ class ArrayCollection implements Collection, Selectable
      * Initializes a new ArrayCollection.
      *
      * @param array $elements
+     *
+     * @psalm-param array<T> $elements
```
weirdan:
```suggestion
     * @psalm-param array<TKey,T> $elements
```

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -31,12 +31,18 @@
...
+ * @template T
+ * @template-implements Collection<T>
+ * @template-implements Selectable<T>
  */
 class ArrayCollection implements Collection, Selectable
 {
     /**
      * An array containing the entries of this collection.
      *
+     * @psalm-var T[]
```
weirdan:
```suggestion
     * @psalm-var array<TKey,T>
```

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -31,12 +31,18 @@
  * and may break when we change the internals in the future. If you need to
  * serialize a collection use {@link toArray()} and reconstruct the collection
  * manually.
+ *
+ * @template TKey of string|int
+ * @template T
+ * @template-implements Collection<T>
+ * @template-implements Selectable<T>
```
weirdan:
```suggestion
 * @template-implements Selectable<TKey,T>
```

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -31,12 +31,18 @@
  * and may break when we change the internals in the future. If you need to
  * serialize a collection use {@link toArray()} and reconstruct the collection
  * manually.
+ *
+ * @template TKey of string|int
+ * @template T
+ * @template-implements Collection<T>
```
weirdan:
```suggestion
 * @template-implements Collection<TKey,T>
```

lib/Doctrine/Common/Collections/AbstractLazyCollection.php
```diff
@@ -6,12 +6,16 @@
...
+ *
+ * @template T
+ * @template-implements Collection<T>
  */
 abstract class AbstractLazyCollection implements Collection
 {
     /**
      * The backed collection to use
      *
+     * @psalm-var Collection<T>
```
weirdan:
```suggestion
     * @psalm-var Collection<TKey,T>
```

lib/Doctrine/Common/Collections/AbstractLazyCollection.php
```diff
@@ -6,12 +6,16 @@
 
 /**
  * Lazy collection that is backed by a concrete collection
+ *
+ * @template T
+ * @template-implements Collection<T>
```
weirdan:
```suggestion
 * @template TKey
 * @template-implements Collection<TKey,T>
```

lib/Doctrine/Common/Collections/Collection.php
nschoellhorn: Why do you want the arguments to be optional? Is there a use case where you want to pass a closure in that doesn't work with either the value or the key?

nschoellhorn:
After another round of improvements (thanks @weirdan), we are now only missing the discussion on the (possibly) optional parameters of the closures for `map()`, `filter()`, etc.

lib/Doctrine/Common/Collections/Collection.php
weirdan: I can easily imagine passing a closure that works with keys only. But if you're able to ignore the values, why shouldn't you be able to omit keys as well? Granted, it's more useful for things like `filter` and `partition`, where you could want random sampling:
```php
$coll->filter(function() { return rand() < 0.1; /* take 10 % of the entries */ });
```
For `map`, you may want to construct the collection having the same keys, but some placeholder or random values:
```php
$item = new Entity;
$coll->map(function() use ($item) { return $item; });
```
`exists` and `forAll` are cases where ignoring both key and values is probably the least useful, so they included mostly for completeness, as it's technically not a problem passing no-args closures to them.

nschoellhorn: Alright, yeah. Makes sense. Gonna add your suggestions 👍

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -32,6 +37,10 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * @param mixed $element The element to add.
      *
      * @return bool Always TRUE.
+     *
+     * @psalm-param T $element
+     * @psalm-return true
+     * @psalm-return true
```
weirdan: Seems like a unintentional duplicate

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -209,6 +260,9 @@ public function map(Closure $func);
      * @return Collection[] An array with two elements. The first element contains the collection
      *                      of elements where the predicate returned TRUE, the second element
      *                      contains the collection of elements where the predicate returned FALSE.
+     *
+     * @psalm-param Closure(TKey, T):bool $p
```
weirdan:
```suggestion
     * @psalm-param Closure(TKey=, T=):bool $p
```
As discussed previously, sometimes it makes sense to partition collection randomly or based on keys only.

Ocramius:
Looks good to me. @doctrine/doctrinecore check please :100:

composer.json
```diff
@@ -17,13 +17,17 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"},
         {"name": "Johannes Schmitt", "email": "schmittjoh@gmail.com"}
     ],
+    "suggest": {
+        "vimeo/psalm": "Psalm allows you to enforce template types for your collections."
```
Majkl578: This is to be dropped - Psalm is not a tool for production, just like PHPStan or Doctrine CS are not.
Feel free to open issue on Composer repo to add `suggest-dev` but it doesn't belong to `suggest`.

Ocramius:
> Primarily I don't like the idea of having proprietary tool-specific annotations in library's public API.

Right now, I see no better alternative to this, but also bo drawbacks, since the metadata is not runtime code.

> This also increases maintenance burden

 > all maintainers will now be required to understand these annotations, understand Psalm, be capable of writing all this and using Psalm.

I don't see the problem with learning the type annotations introduced so far: it is a sensible set of type declarations that we should familiarise with (... continues on the precedent bit).

 > Overall it sounds like a dangerous precedent to me

I've asked @nschoellhorn to do this precisely to set a precedent. `doctrine/collections` is a very high impact, very low risk, very high visibility library, used in thousands of projects. Adding type declarations to it allows those projects to benefit from these additions immediately, without risky runtime changes.

> random proprietary annotations/hacks

Besides my opinion that plsam is doing it right here, I think we set things in motion almost a decade ago with `doctrine/annotations` there, effectively denying our ability to state "random annotation hacks".

> and endorsing Psalm

I don't see a problem with this either: we did pish  phpstan a lot (heck, I even actively sponsored it with money out of my pocket), but psalm is indeed much more useful here, so the endorsement is very much earned. Writing library-specific inference plugins is surely not the way to go either, as this patch demonstrates that a flexible and powerful DSL provides much better results without runtime code additions.

> This also increases maintenance burden

I don't mind asking @nschoellhorn to join the team and help us out here: he jas been working very diligently on a couple releases for my own packages, and demonstrated to be working with extreme care, so if the maintenance and learning burden is an issue, let's onboard folks that will help us out in getting better on this front.

> through some XML metadata format

Must disagree here as well, because of a few added issues:

 * Much more complex DSL: targeted annotations make the coupling between type declarations and code much more explicit. I'm a fan of XML, but here we are really defining static types that are part of the code definitions.
 * Distribution - now we need to teach downstream consumers about how to integrate those, while with this patch/approach there are no downstream code changes

I'd also add that we (PHP developers) have wanted generics for quite a while: well here they are, and in a quite harmless and BC compliant format too! Let's use them, I'd say.

composer.json
Ocramius: Agreed: this will end up just being noise

ondrejmirtes:
I expect PHPStan to learn and take advantage those annotations in upcoming months too :)  👍 (I'm even thinking about reading those ` @psalm-return` and `@psalm-param` because why not :)

Ocramius:
@ondrejmirtes that would be aaaaaaaaaaawesome

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -32,6 +37,9 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * @param mixed $element The element to add.
      *
      * @return bool Always TRUE.
+     *
+     * @psalm-param T $element
+     * @psalm-return true
```
alcaeus: Is psalm-return necessary when it’s the same as the return annotation?

weirdan: It's not the same, `@psalm-return` here specifies narrower type (`true`) as opposed to `bool` specified by `@return`.

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -396,7 +409,9 @@ public function matching(Criteria $criteria)
                 $next = ClosureExpressionVisitor::sortByField($field, $ordering === Criteria::DESC ? -1 : 1, $next);
             }
 
-            uasort($filtered, $next);
+            if ($next) {
+                uasort($filtered, $next);
```
ostrolucky: I would like to actually see reason why was this change sneaked here

Ocramius: According to the loop above, `$next` may be `null` from a type perspective. That is impossible to reach, assuming `ClosureExpressionVisitor::sortByField()` is not nullable

weirdan: I think Psalm should actually be able to figure that out, given that it's set in the loop over non-empty `$orderings`. I'm going to file Psalm bug for this.

weirdan: Ticketed under vimeo/psalm#1470.

lib/Doctrine/Common/Collections/Collection.php
alcaeus: I believe we should just change the return type itself then.

weirdan: Other tools may not support literal boolean types like `true` and `false`.

nschoellhorn: According to the [FIG specification](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#appendix-a-types) it should indeed be possible to use literal types in the normal PHPDoc. Hence, I agree that we could remove the psalm annotation and modify the other annotation to state `true` as the return type. If other tools don't support that, they seemingly don't implement the spec correctly as far as I can tell.

lib/Doctrine/Common/Collections/ArrayCollection.php
muglug: Thanks, I've fixed (but it'll be a few days yet before I release).

lib/Doctrine/Common/Collections/Collection.php
weirdan: Ah, I wasn't aware that was a part of PSR-5 (even though it's not a formally ratified standard, it's still widely used). Then I guess it'd be ok to replace it with `@return true`.

muglug:
Firstly @ondrejmirtes that's terrific news!

These non-PHPDoc annotations are necessary for consumers to understand how the code works, and also to help them avoid bugs.

Here's some buggy code with valid PHPDoc types:

```php
<?php declare(strict_types=1);

/** @param string[] $arr */
function foo(array $arr) : void {
  foreach ($arr as $k => $v) {
    strlen($k);
  }
}

foo(["hello", "goodbye"]);
```

To catch this particular bug you have to use a non-PHPDoc-compliant annotation:

```php
<?php declare(strict_types=1);

/** @param array<string, string> $arr */
function foo(array $arr) : void {
  foreach ($arr as $k => $v) {
    strlen($k);
  }
}

foo(["hello", "goodbye"]);
```

Everything Psalm has done, Phan (which originated `@template`, designed by PHP's creator), and PHPStan has done to improve PHP's type system follows on from that example.

Types document the interface between functions. Making types more specific makes for more concise code, and improves its type safety.

The tool-specific annotations above (which won't be tool-specific when PHPStan supports them) make the interface between functions less error-prone, and correct errors made in the evolution of PHPDoc.

---

In 2013 PhpStorm didn't support using the semantically accurate `ArrayObject<TKey, TValue>`, but there was a workaround - if you had a union of `ArrayObject|TValue[]` you could use autocompletion on `TValue` within `foreach` loops:

```php
/** @param ArrayObject|MyClass[] */
function bar(array $arr) {
  foreach ($arr as $obj) {
    $obj-> // autocompletion works here for MyClass
  }
}
```

Developers love autocompletion, so they started using this annotation in a lot of places. PHPDoc was updated in 2013 to say, essentially, "if you want autocompletion in your IDE, write your code this way".

Fast forward six years, and many developers think this is the One True Way to annotate `ArrayObject<TKey, TValue>` (Psalm supports the format, but behind a config flag). I am sure that if PhpStorm had opted instead to support `ArrayObject<int|string, MyClass>`, and had popularised the use of a generic annotations, PHPDoc would support them and none of us would be having this discussion. 

---

Psalm (made at and for Vimeo) supports these more-specific type annotations because their use has repeatedly prevented embarrassing accidents on Vimeo's website, and in our API.

If Psalm, Phan or PHPStan can use the annotations in this PR to prevent other similar accidents around the globe, our end users will silently thank us.

👍 from me

nschoellhorn:
I'm sorry to dismiss your reviews again. Also, I think we need to wait for the fix from @muglug (Psalm) first and then I need to bump the dependency of Psalm once again. I can also roll back the change to the actual code regarding `uasort()` then.

muglug:
@nschoellhorn I've just released 3.2.2 with the bugfix

Ocramius:
@nschoellhorn please also rebase

nschoellhorn:
Thank you @muglug for the fast update once again. I've bumped the dependency of Psalm to the new version and rolled back the change regarding `uasort()`. No issues anymore.

Ocramius:
Yes, type checks propagate statically

SenseException:
Should we have some kind of tests that prevent logical changes in the psalm annotations of doctrine/collections?

Ocramius:
That would be like `roave/backwards-compatibility-check`: the ecosystem doesn't have that kind of stuff yet (even BC checker is way too new)

weirdan:
> Should we have some kind of tests that prevent logical changes in the psalm annotations of doctrine/collections?

You can take a look at https://github.com/weirdan/doctrine-psalm-plugin/blob/master/tests/acceptance/Collections.feature - most of those tests would apply here as well. Feel free to borrow them.

nschoellhorn:
Looks good, I can add that if wanted.

Ocramius:
Let's please defer that for now

Ocramius:
Or better: separate patch (based on this one)

nschoellhorn:
Sure, I will prepare one soon then.

Ocramius:
Right, calling the merge here :ship:

PR #179, opened 2019-03-24
Enforce no BC breaks against v1.5.0

Ocramius:
Possibly a good idea for branches in `1.x`, not to be merged into `master`? Just throwing it out there: this would prevent most feedback along the lines of "we can't do that because BC" from human-to-human, which is frustrating.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,3 +52,8 @@ jobs:
       env: STATIC_ANALYSIS
       install: travis_retry composer install --prefer-dist
       script: vendor/bin/psalm
+
+    - stage: Backwards Compatibility
+      env: STATIC_ANALYSIS
+      install: travis_retry composer install --prefer-dist
+      script: vendor/bin/roave-backward-compatibility-check --from=v1.5.0
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
         "phpunit/phpunit": "^7.0",
         "doctrine/coding-standard": "^6.0",
         "phpstan/phpstan-shim": "^0.9.2",
+        "roave/backward-compatibility-check": "^2.1.0",
         "vimeo/psalm": "^3.2.2"
     },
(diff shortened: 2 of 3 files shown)


greg0ire:
Looks like it cannot be done because the lib requires php 7.2 and this package is supposed to be compatible with 7.1

Ocramius:
Can bump the PHP version, if needed for `1.7.x`. The idea is that `master` is open for BC breaks, while `1.x` is not.

greg0ire:
I like the idea! Regarding your first sentence, did you mean `1.6.x` ? Why not bump the php version when releasing `1.6.0`?

.travis.yml
```diff
@@ -52,3 +52,8 @@ jobs:
       env: STATIC_ANALYSIS
       install: travis_retry composer install --prefer-dist
       script: vendor/bin/psalm
+
+    - stage: Backwards Compatibility
+      env: STATIC_ANALYSIS
+      install: travis_retry composer install --prefer-dist
+      script: vendor/bin/roave-backward-compatibility-check --from=v1.5.0
```
Majkl578: Hardcoding tag name here is not a good idea, can we use some kind of wildcard, like _latest 1.x tag_?

Majkl578 (changes requested):
Bumping library requirement to PHP ^7.2 just to satisfy a dev-dependency sounds like a bit too much IMO. We should seek other solution.

ostrolucky:
I agree with not bumping requirement because of dev dependency

mpdude:
Is this still relevant, or could this PR be closed?

Ocramius:
IMO relevant, but the tooling has evolved, and there's no need to specify a base branch anymore, IIRC.

I won't have time to pursue this, but usage via `tools/composer.json` + `tools/composer.lock` is endorsed.

PR #185, opened 2019-03-25
Update Collection::filter docblock to add key parameter to closure

nicholasruunu:
Since the key was added to the closure of `ArrayCollection::filter` in 1.6.0 I guess this needs to be updated.

### DIFF
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -221,7 +221,7 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      *
      * @return Collection A collection with the results of the filter operation.
      *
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -221,7 +221,7 @@ public function exists(Closure $p);
      *
      * @return Collection A collection with the results of the filter operation.
      *
-     * @psalm-param Closure(T=):bool $p
+     * @psalm-param Closure(T=, TKey):bool $p
```
Ocramius: Hmm, optionality should be clarified here too

nicholasruunu: Fixed.

PR #186, opened 2019-03-25
Use @psalm-template annotation to avoid clashes

muglug:
Following reports that some users encountered issues with `@template` annotations, this PR replaces `@template` with the functionally-equivalent `@psalm-template`.

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -7,8 +7,8 @@ use Closure;
 /**
  * Lazy collection that is backed by a concrete collection
  *
- * @template TKey of array-key
- * @template T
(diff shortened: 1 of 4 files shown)


Ocramius:
Verified via following script:

```php
<?php

declare(strict_types=1);

namespace A;

use Doctrine\Common\Annotations\AnnotationReader;
use Doctrine\Common\Collections\Collection;

(function () {
    require_once __DIR__ . '/vendor/autoload.php';

    AnnotationReader::addGlobalIgnoredName('psalm');

    $reader = new AnnotationReader();

    var_dump($reader->getClassAnnotations(new \ReflectionClass(Collection::class)));
})();
```

Produces:

```sh
php check-patch-186.php 
array(0) {
}
```

Ocramius:
Backported to `1.6` via 688aef66f88911572108ac87ca2445787332e35b

PR #187, opened 2019-03-26
Documentation for serialization of collections

SenseException:
Hints about the serialization of collections can be found in the collection code comments, but they are missing in the documentation. This PR contains a minimal documentation about the serialization.

Do we need more examples or a section for third-party serialization libraries?

### DIFF
--- a/docs/en/serialization.rst
+++ b/docs/en/serialization.rst
@@ -1,0 +1,29 @@
+Serialization
+=============
+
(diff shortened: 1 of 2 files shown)


Ocramius:
> Do we need more examples or a section for third-party serialization libraries?

I'd avoid that for now - this kind of exposed/documented API is sufficient :+1:

PR #192, opened 2019-04-25
[2.0] Strict types, native parameter & return types 

jwage:
Work originally started in #125 

- Bump dev-master branch alias to `2.0.x-dev`
- Require PHP 7.2
- Upgrade PHPStan to 0.11
- Increase PHPStan to Level 7
- Add strict type hints everywhere
- Remove some of the exclusion rules in `phpcs.xml.dist`
- Sort packages in `composer.json`

### DIFF
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -119,8 +119,8 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
     /**
      * Gets all keys/indices of the collection.
      *
-     * @return int[]|string[] The keys/indices of the collection, in the order of the corresponding
-     *               elements in the collection.
+     * @return array<int, int|string> The keys/indices of the collection, in the order of the corresponding
+     *                                elements in the collection.
      *
      * @psalm-return TKey[]
      */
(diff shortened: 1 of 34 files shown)


stof:
should we add psalm annotations for generics for the offset types ?

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -292,5 +291,30 @@ public function indexOf($element);
...
      */
-    public function slice($offset, $length = null);
+    public function slice(int $offset, ?int $length = null) : array;
+
+    public function count() : int;
+
+    public function getIterator() : Traversable;
+
+    /**
+     * @param mixed $offset
```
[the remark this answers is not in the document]
jwage: It originally had the inheritDoc tag but psalm/phpstan didn't like it so I was trying different things trying to get it to work. I will change it to `string|int`.

jwage:
@stof I am new to psalm. What does that look like? Do we have any examples in Doctrine already I can look at?

stof:
we do have enhanced psalm annotations in the Collection interface, making generic

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -292,5 +291,32 @@ public function indexOf($element);
      *
      * @psalm-return array<TKey,T>
      */
-    public function slice($offset, $length = null);
+    public function slice(int $offset, ?int $length = null) : array;
+
+    public function count() : int;
```
ostrolucky: I'm 👎 on copying core interfaces just to add type hints

jwage: This was already in the PR when I took it over and frankly I didn't understand why it was there since those interfaces are already defined upstream in PHP. I will remove them.

PR #196, opened 2019-04-30
Change Collection::slice() method to return an instance of Collection

### LINKED ISSUE #195: slice() method should return a Collection instead of an array
See question here where this was originally requested https://github.com/doctrine/collections/pull/125#issuecomment-378990965

jwage:
Fixes #195

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -254,7 +254,7 @@ abstract class AbstractLazyCollection implements Collection
     /**
      * {@inheritDoc}
      */
-    public function slice(int $offset, ?int $length = null): array
+    public function slice(int $offset, ?int $length = null): Collection
     {
         $this->initialize();
 
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -290,5 +290,5 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * @return mixed[]
      * @psalm-return array<TKey,T>
      */
-    public function slice(int $offset, ?int $length = null): array;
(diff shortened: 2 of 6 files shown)


lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -286,9 +286,9 @@ public function indexOf($element);
      * @param int      $offset The offset to start from.
      * @param int|null $length The maximum number of elements to return, or null for no limit.
      *
-     * @return array
+     * @return Collection
      *
-     * @psalm-return array<TKey,T>
+     * @psalm-return Collection<TKey, T>
```
greg0ire: Should we add `@return static` here, and then change the type hints of implementors to `self` when dropping php 7.3?

jwage: I am not sure. Can that be done in another issue/PR?

lib/Doctrine/Common/Collections/AbstractLazyCollection.php
```diff
@@ -268,7 +268,7 @@ public function indexOf($element)
     /**
      * {@inheritDoc}
      */
-    public function slice(int $offset, ?int $length = null) : array
+    public function slice(int $offset, ?int $length = null) : Collection
```
Ocramius: What is the impact of this BC break on the OSS ecosystem? Is `->slice()` called often?

jwage: I am not sure. Trying to think of a way to search for that on github somehow.

lib/Doctrine/Common/Collections/Collection.php
greg0ire: Sure! It's a tiny detail, definitely not a blocker

greg0ire:
Oh I just noticed this has been taken over already.

Closing in favor of https://github.com/doctrine/collections/pull/264

PR #201, opened 2019-06-08
Return `static` for `ArrayCollection::createFrom()` and `ArrayCollection#filter()`

### LINKED ISSUE #200: Psalm return type when subclassing
I have some collections subclassing `ArrayCollection` and doing thing like this:

```php
class EpisodeCollection extends ArrayCollection
{
    public function getInPeriod(Period $period): self
    {
        return $this->filter(static function (EpisodeInterface $episode) use ($period): bool {
            return $period->contains($episode->getDate());
        });
    }
```

From my perspective this works fine, I want to return an instance of my own collection from its method and that's how upstream `filter()` does work when subclassed (down to `createFrom()` returning `new static`).

However this gets Psalm all bothered:

```
ERROR: MoreSpecificReturnType - src\Tv\Collection\EpisodeCollection.php:12:50 - The declared return type 'Rarst\Tv\Collection\EpisodeCollection' for Rarst\Tv\Collection\EpisodeCollection::getInPeriod is more specific than the inferred return type 'Doctrine\Common\Collections\ArrayCollection<array-key, mixed>'
    public function getInPeriod(Period $period): self


ERROR: LessSpecificReturnStatement - src\Tv\Collection\EpisodeCollection.php:14:16 - The type 'Doctrine\Common\Collections\ArrayCollection<array-key, mixed>' is more general than the declared return type 'Rarst\Tv\Collection\EpisodeCollection' for Rarst\Tv\Collection\EpisodeCollection::getInPeriod
        return $this->filter(function (EpisodeInterface $episode) use ($period): bool {
            return $period->contains($episode->getDate());
        });
```

I am guessing this is over upstream methods being `@psalm-return ArrayCollection<TKey,T>` ?

Am I missing something here on how I am (not) supposed to subclass and/or document this?

muglug:
Fixes #200

### DIFF
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -86,7 +86,7 @@ class ArrayCollection implements Collection, Selectable
      * @return static
      *
      * @psalm-param array<TKey,T> $elements
-     * @psalm-return ArrayCollection<TKey,T>
+     * @psalm-return static<TKey,T>
      */
     protected function createFrom(array $elements)
     {
@@ -323,7 +323,7 @@ class ArrayCollection implements Collection, Selectable
      *
      * @return static
      *
-     * @psalm-return ArrayCollection<TKey,T>
+     * @psalm-return static<TKey,T>
      */
     public function filter(Closure $p)
     {


greg0ire (approved):

greg0ire:
I think you should target 1.6

muglug:
Yeah, this has no breaking changes - it just improves type information

greg0ire:
Please rebase then.

greg0ire:
Ah I can do it for you :)

bramcordie:
I already had a discussion over at the PHPStan project: https://github.com/phpstan/phpstan/discussions/4825#discussioncomment-583813.

What it seems to boil down to is that `@psalm-return static<TKey,T>` does not make sense and `@return static` is sufficient? Was the `<TKey,T>` left over from changing `ArrayCollection` to `static` or is this intentional?

Issue #205, opened 2019-07-07
Get default value if target element not found


spam312sn:
Update `get` method with one more additional method parameter which will be used as default value if target element couldn't be found.

Example: [Symfony ParameterBag][1]

[1]: https://github.com/symfony/http-foundation/blob/9a96d77ceb1fd913c9d4a89e8a7e1be87604be8a/ParameterBag.php#L74-L85

someniatko:
personally i think `$collection->get($id) ?? $default` is more expressive, than `$collection->get($id, $default)`

malarzm:
The `$collection->get($id) ?? $default` approach won't work if you need to store `null` in the collection however I'm not sure a lot of people are storing `null` in the collections.

SenseException:
A change of the `get()` method API would break packages where collection is a dependency and still uses the variant without a default value argument. Projects are implementing the `Collection` interface and this would also affect Doctrine projects like ORM and the `PersistentCollection`. At the current time Packagist lists 628 packages that are dependents.

The influence and circumstances of such a change on maintenance compared to its benefit doesn't pay off IMHO.

alcaeus:
While the BC break is definitely to be avoided, we'll have to think about how to solve this in the future: there were numerous feature requests that would've been nice to implement but were skipped because of the BC implications.

jwillp:
Since the `get` method returns null in case the value is not found (which serves as some kind of a default), wouldn't the following signature `get($key, $default = null)` prevent BC?

someniatko:
@jwillp No, it will not. BC does not only concern interface usage, but also interface implementation.
Changing method signature would mean each `CollectionInterface` implementor would need to support the second parameter.

PR #206, opened 2019-07-07
#205, Get default value if no target element found

spam312sn:
Return defined default value if no target element found or null if default value is not set


SenseException:
Thank you for your contribution.

I'm not sure about this one, because I considered the default value something that belongs to the project, that is using the collection. The user checks if a key exists with `Collection::containsKey()` and then handles an error or uses the  fetched value.

If this is considered to be merged, all other Doctrine projects need to adapt to the interface changes.

PR #211, opened 2019-10-28
fix #210 by including psalm-* annotations to ArrayCollection#map()

### LINKED ISSUE #200: Psalm return type when subclassing

### LINKED ISSUE #210: Invalid psalm type inference for ArrayCollection#map()
Similar to #200 but about `map()` method.

I have created a reproduction repository: https://github.com/someniatko/doctrine-collections-map-psalm

doctrine/collections: v1.6.2
vimeo/psalm: 3.6.2

someniatko:
Apparently, psalm cannot reuse `@psalm-template` from parent's method declaration using `@inheritdoc`, so we cannot just specify `@psalm-return static<TKey, U>`.

Fixes #210

### DIFF
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -312,6 +312,10 @@ class ArrayCollection implements Collection, Selectable
      * {@inheritDoc}
      *
      * @return static
+     *
+     * @psalm-template U
+     * @psalm-param Closure(T=):U $func
+     * @psalm-return static<TKey, U>
      */
     public function map(Closure $func)
     {


Ocramius:
@muglug can I please get your opinion, before we go ahead and repeat docblocks in inheritance trees

muglug:
Looks good to me!

muglug:
Psalm sees `@return static` and that always takes precedence over something inherited. This PR will fix the return type.

Ocramius:
Thanks @someniatko!

someniatko:
@muglug @Ocramius I have no idea why, but this issue still reproduces itself on https://github.com/someniatko/doctrine-collections-map-psalm

I found out that if in
```php
    /**
     * @return static
     *
     * @psalm-template U
     * @psalm-return static<TKey, U>
     */
    public function map(Closure $func)
```

I replace `static` with `ArrayCollection` in `psalm-return` annotation, psalm does not longer find errors.
I believe this is a bug in psalm.

psalm 3.6.4
doctrine/collections 1.6.3

PR #216, opened 2019-11-13
Support using embeddables in criterias with ArrayCollection

garak:
This is a re-proposal of PR #27 by @mnapoli


mnapoli:
I wanted to stop being pinged on the latest PR and now I just got 4 new pings this morning…

garak:
@mnapoli I'm sorry you feel bothered. I guess github is already providing tools to avoid notifications.

garak:
I know it's not nice to bump, but two months passed and I'm still waiting for a review.

greg0ire:
You are targetting master, is it because there is a BC-break in there? Are dots already supported?  What is the current behavior regarding them?

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -37,6 +38,13 @@ class ClosureExpressionVisitor extends ExpressionVisitor
      */
     public static function getObjectFieldValue($object, string $field)
     {
+        if (strpos($field, '.') !== false) {
+            [$field, $subField] = explode('.', $field, 2);
```
greg0ire: What should happen if there is more than one dot?

garak:
> You are targeting master, is it because there is a BC-break in there? Are dots already supported? What is the current behavior regarding them?

I follow "everything to master first" rule.
Dots are currently not supported and this PR aims to add such support

greg0ire:
> I follow "everything to master first" rule.

Just checking, we probably need to change the default branch here.
Since your PR seems BC, please target 1.6.x

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
garak: I guess it would just won't work (just like it's not working without this PR)

garak:
> > I follow "everything to master first" rule.
> 
> Just checking, we probably need to change the default branch here.
> Since your PR seems BC, please target 1.6.x

Tried to, but it's a mess. Tried to rebase, with much success. Do you think it would be better to re-open a new PR with relevant changes?

> > Dots are currently not supported and this PR aims to add such support
> 
> So there is a crash?

I guess it would just don't work.

greg0ire:
I'm going to rebase it interactively for you.

greg0ire:
Here are roughly the steps I'm going to follow FYI:

1. Fetch all new commits: `git fetch --all`. ⬇️ 
2. Rebase on what you fetched interactively: ✂️ 
    1. `git rebase -i origin/1.6.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin repository.
    2. A window will show up with many lines, remove every line but the last one, which should correspond to your commit.
   2. If you run into a conflict: 💥 
        1. fix it with `git mergetool`. 😷 
        1. continue on your merry way with `git rebase --continue`. ⏩ 
3. Force push to overwrite all this : `git push --force`. ⬆️

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
greg0ire: Can you describe "not working" more precisely? I want to know if the error message people are getting is clear enough right now.

garak: See issue https://github.com/doctrine/orm/issues/3783

greg0ire: I think it looks good enough 👍

greg0ire (dismissed):

greg0ire:
Are there any docs that need to be updated?

garak:
> Are there any docs that need to be updated?

I took a look to current docs but I couldn't find a spot to possibly add something relevant.
This is just a fix for something that is expect to work, but did not

greg0ire:
Thanks for checking!

greg0ire:
This is a feature, so I'm going to retarget to 1.7.x

greg0ire (dismissed):
That comment was automated, both branches correspond to the same commit so my approval stays.

garak:
I just solved conflicts

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @garak !

PR #217, opened 2019-11-13
Use variadic params in ExpressionBuilder#andX() and #orX()

someniatko:
### What?
Changes `andX()` and `orX()` method declarations of `ExpressionBuilder` class from `andX($x = null)` to `andX(...$expressions)`.

### Why?
When using psalm static analyzer, it may report a false positive "TooManyArguments" error when making a call to `ExpressionBuilder#andX()` (`orX()`). This is because in its declaration there is only one parameter.

### Can it be done?
As long as doctrine/collections `^1.6` depend on PHP `^7.1` , and variadic params were introduced in PHP 5.6, this can be done in this library. It does also not introduce any BC breaks.

### DIFF
--- a/lib/Doctrine/Common/Collections/ExpressionBuilder.php
+++ b/lib/Doctrine/Common/Collections/ExpressionBuilder.php
@@ -5,7 +5,6 @@ namespace Doctrine\Common\Collections;
 use Doctrine\Common\Collections\Expr\Comparison;
 use Doctrine\Common\Collections\Expr\CompositeExpression;
 use Doctrine\Common\Collections\Expr\Value;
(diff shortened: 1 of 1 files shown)


alcaeus:
> It does also not introduce any BC breaks.

Except it does: https://3v4l.org/psZ74

Sorry, can't merge this as the signature must not change.

someniatko:
@alcaleus maybe it makes sense to change just PHPDoc then?

alcaeus:
Yes, updating PHPDoc to show the method as variadic makes sense.

someniatko:
@alcaeus can I create similar MR to master?

PR #218, opened 2019-11-13
reflect that ExpressionBuilder#andX() and orX() params are variadic in PHPDoc

someniatko:
Like https://github.com/doctrine/collections/pull/217 but changes only PHPDoc

### DIFF
--- a/lib/Doctrine/Common/Collections/ExpressionBuilder.php
+++ b/lib/Doctrine/Common/Collections/ExpressionBuilder.php
@@ -17,7 +17,7 @@ use function func_get_args;
 class ExpressionBuilder
 {
     /**
-     * @param mixed $x
+     * @param mixed ...$x
      *
      * @return CompositeExpression
      */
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Collections/ExpressionBuilder.php
```diff
@@ -17,7 +17,7 @@
 class ExpressionBuilder
 {
     /**
-     * @param mixed $x
+     * @param mixed ...$x
```
Ocramius: This syntax is not really understood by current tooling 🤔

someniatko: https://github.com/vimeo/psalm/blob/master/tests/VariadicTest.php

Ocramius: Oh, that's very good news!

someniatko:
force-pushed to change commit author

alcaeus:
@Ocramius sometimes GitHub and Travis-ci just need a gentle kick in the right direction ;)

PR #220, opened 2019-11-14
Allow for covariant elements in Collection

pamil:
With covariant return types in PHP 7.4, I think Collection should also allow for that.

Getting this errors with Psalm and covariant elements:

 - `
ERROR: InvalidReturnType - src/Sylius/Component/Core/Model/Promotion.php:41:36 - The declared return type 'Doctrine\Common\Collections\Collection<array-key, Sylius\Component\Channel\Model\ChannelInterface>' for Sylius\Component\Core\Model\Promotion::getChannels is incorrect, got 'Doctrine\Common\Collections\Collection<array-key, Sylius\Component\Core\Model\ChannelInterface>'
    public function getChannels(): Collection
`

 - `
ERROR: InvalidReturnStatement - src/Sylius/Component/Core/Model/Promotion.php:43:16 - The type 'Doctrine\Common\Collections\Collection<array-key, Sylius\Component\Core\Model\ChannelInterface>' does not match the declared return type 'Doctrine\Common\Collections\Collection<array-key, Sylius\Component\Channel\Model\ChannelInterface>' for Sylius\Component\Core\Model\Promotion::getChannels
        return $this->channels;
`

To provide some background:
 - `Sylius\Component\Core\Model\ChannelInterface` extends `Sylius\Component\Channel\Model\ChannelInterface`
 - `Sylius\Component\Core\Model\Promotion` extends `Sylius\Component\Promotion\Model\Promotion`
 - `Sylius\Component\Promotion\Model\Promotion::getChannels()` defines its return type as `Collection<array-key, Sylius\Component\Channel\Model\ChannelInterface>`
 - `Sylius\Component\Core\Model\Promotion::getChannels()` returns property typed with `Collection<array-key, Sylius\Component\Core\Model\ChannelInterface>`

Please let me know if it should be treated as a bugfix or as a new feature, I can rebase if needed.

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -7,8 +7,8 @@ use Closure;
 /**
  * Lazy collection that is backed by a concrete collection
  *
- * @psalm-template TKey of array-key
- * @psalm-template T
+ * @psalm-template-covariant TKey of array-key
+ * @psalm-template-covariant T
  * @template-implements Collection<TKey,T>
  */
 abstract class AbstractLazyCollection implements Collection
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -32,8 +32,8 @@ use function uasort;
  * serialize a collection use {@link toArray()} and reconstruct the collection
  * manually.
  *
- * @psalm-template TKey of array-key
- * @psalm-template T
+ * @psalm-template-covariant TKey of array-key
+ * @psalm-template-covariant T
  * @template-implements Collection<TKey,T>
  * @template-implements Selectable<TKey,T>
  */
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -24,8 +24,8 @@ use IteratorAggregate;
  * position unless you explicitly positioned it before. Prefer iteration with
  * external iterators.
  *
- * @psalm-template TKey of array-key
- * @psalm-template T
+ * @psalm-template-covariant TKey of array-key
+ * @psalm-template-covariant T
  * @template-extends IteratorAggregate<TKey, T>
  * @template-extends ArrayAccess<TKey|null, T>
  */
--- a/lib/Doctrine/Common/Collections/Selectable.php
+++ b/lib/Doctrine/Common/Collections/Selectable.php
@@ -14,8 +14,8 @@ namespace Doctrine\Common\Collections;
  * this API can implement efficient database access without having to ask the
  * EntityManager or Repositories.
  *
- * @psalm-template TKey as array-key
- * @psalm-template T
+ * @psalm-template-covariant TKey as array-key
+ * @psalm-template-covariant T
  */
 interface Selectable
 {


pamil:
Reopening to trigger a build based on 1.6.x branch.

pamil:
Psalm needs to be updated first, the current version does not support `@psalm-template-covariant`.

someniatko:
@pamil You cannot use covariance in Collection. Consider this case:

```php

interface Animal {}
class Dog implements Animal {}
class Cat implements Animal {}

/** @param Collection<Animal> $col */
function addDog(Collection $col): void
{
    $col->add(new Dog());
}

/** @var Collection<Cat> $cats */
$cats = new ArrayCollection([ new Cat, new Cat ]);
addDog($cat);
```

alcaeus:
@someniatko your example is correct if this was about method arguments, but this is about return types, where covariance is allowed. Without collections:

```
interface foo {
    public function doSomething(): foo;
}

class bar implements foo {
    public function doSomething(): bar {};
}
```

This is perfectly fine, and should work the same way with collections:
```
interface Animal {}
class Dog implements Animal {}

interface Shelter {
    public function getAnimals(): Collection<Animal>;
}

class DogShelter implements Shelter {
    public function getAnimals(): Collection<Dog>;
}
```

This second example is theoretically valid, but not supported with the current annotations. This PR aims to change that.

someniatko:
@alcaeus You're right about return types which actually can be covariant. However by marking whole Collection interface's generic param covariant, you allow this param to be covariant in both Collection's return types, and it's methods' params, e.g. `add`. There is no way to mark only certain methods to be covariant. You should then separate the Collection into two, one of which will contain only those methods which do not use `T` as method param.

alcaeus:
Ah, I see. In that case, this isn't what we want. @muglug any way we can achieve what Pamil is trying to do here?

pamil:
You're both right, thanks a lot for your input 🎉

Ocramius:
This isn't really fixable, because the collection API allows for both **writes** and **reads** which require in/out values to be invariant between each other.

Therefore consumers relying on `Collection<T>` must also require invariant types.

If we had two different interfaces for reading and for writing, then variance would be possible in write-only-interface consumers or read-only-interface consumers.

someniatko:
@Ocramius separating reads and writes also won't solve the issue, consider method `contains(T $item)`

akondas:
How, then, to solve such a case?
Let's assume then that I have a policy collection and `Policy` is an interface:
```php
interface Policy {}

/** @var Collection<Policy> $collection */
```
How can I check if a collection has a given policy when I cannot allow for covariance?
```php
class SpecificPolicy implements Policy {}

$collection->contains(new SpecificPolicy());
```

Currently psalm permits it but phpstan forbids it (https://phpstan.org/r/da200c51-a972-496c-9db7-858f7f5a979e)

akondas:
Here is a better example: https://psalm.dev/r/46d572ff3b

someniatko:
@akondas in your example, replacing `@template` with `@template-covariant` fixes the issue
https://psalm.dev/r/11bb1d43b4

akondas:
@someniatko yes, for psalm, but phpstan will complain then for methods like `contains`: https://phpstan.org/r/e176b986-6efc-4e0f-9bf4-1766606e8c7e (thanks for response)

PR #221, opened 2019-11-15
Add ArrayCollection map() method capability to access inner array key

didix16:
This enhancement gives to map() Closure the capability to access also the keys of the inner array holding the elements. Until now, we were able to access just the element on map Closure but not the key. So I propose to add the ability to access also the key for do some statements inside the Closure with key related.

I have the need to add it since in my project I must get the asociative key to take a decision in function of the current key element and I thought that may be usefull for everyone in some other projects to access the key of current processing element

### DIFF
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -325,12 +325,12 @@ class ArrayCollection implements Collection, Selectable
      * @return static
      *
      * @psalm-template U
-     * @psalm-param Closure(T=):U $func
+     * @psalm-param Closure(T=, TKey=):U $func
      * @psalm-return static<TKey, U>
      */
     public function map(Closure $func) : Collection
     {
-        return $this->createFrom(array_map($func, $this->elements));
+        return $this->createFrom(array_map($func, $this->elements, array_keys($this->elements)));
     }
 
     /**


lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -325,12 +325,12 @@ public function getIterator() : Traversable
      * @return static
      *
      * @psalm-template U
-     * @psalm-param Closure(T=):U $func
+     * @psalm-param Closure(T=, TKey):U $func
```
Ocramius: `TKey=`, since the parameter is optional

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -325,12 +325,12 @@ public function getIterator() : Traversable
...
      * @psalm-return static<TKey, U>
      */
     public function map(Closure $func) : Collection
     {
-        return $this->createFrom(array_map($func, $this->elements));
+        return $this->createFrom(array_map($func, $this->elements, array_keys($this->elements)));
```
Ocramius: Seems very much valid to me, but requires test additions to prevent regressions later on.

Ocramius (changes requested):
Please adjust commit message (to explain what is being done), PR title and tests (required)

someniatko:
I think you should also update PHPDoc for the `Collection` interface.

alcaeus:
I don't think we should update the documentation for the `Collection` interface: not all collection will implement this functionality that way.

someniatko:
I always thought that ArrayCollection is just an implementation of original Collection interface, and in general it shouldn't add more functionality than declared by the original interface.

Ocramius:
There's a BC problem here if we move it to `Collection`. There's a type problem if we don't move it to `Collection`

Specifically, the signature is currently something like `Collection#map(callable (TVal=) : V) : Collection<TKey, V>`.

If we change it to `Collection#map(callable (TVal=, TKey=) : V) : Collection<TKey, V>`, then consumers are still free to ignore the second callable parameter (undefined behavior).

If we don't change it, then `Collection#map(callable (TVal=) : V) : Collection<TKey, V>` is explicitly incompatible with calling `#map` with `callable (Foo, Bar) : V` as parameter, since this callable is of a stricter (type) than what `Collection#map()` allows.

I don't have a solution for now, but I hope this explains the problem.

someniatko:
> There's a BC problem 

This PR targets master so it shouldn't be a problem afaik.

lib/Doctrine/Common/Collections/ArrayCollection.php
didix16: Oh I forgot to add it. Thanks :+1:

didix16 (commented):
Fixed PHPDoc misstype

lib/Doctrine/Common/Collections/ArrayCollection.php
didix16: :+1 done

greg0ire:
Please format your commit message according to our rules : 

- The commit message subject must be less than 50 characters and tell us *what* you did.
- The commit message body should tell us *why* you did it. It is optional but highly recommended.

Bad example :

```
Fixed bug #989 by removing call to defraculate()
```

Good example:

```
Remove call to defraculate()

Calling this function caused a bug because it interferes 
with calls to getSchmeckles().
Fixes #989
```

More details in CONTRIBUTING.md

didix16:
> Please improve your commit message according to [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches).
> 
> Please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.
> 
> How to do that?
> 
> 1. `git rebase -i origin/master`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
> 2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
> 3. Close your editor, git should do its magic, and you should end up with one commit
> 4. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

Oh sorry. This is my first PR. I never do it before ( Im very newbie ). I just do it the modifications at Github, not in my local, so I don't know if I can do the rebase at the web page.

What should I do now? I have to clone the repository then?

P.S: For future commits and PR I'll keep in my mind the procedure to make a proper PR.

Thanks in advance

greg0ire:
I can totally do it for you if you don't feel comfortable jumping through the following hoops:

```
# Clone the repo
git clone git@github.com:doctrine/collections.git

# Enter it
cd collections

# Add your fork as a remote
git remote add fork git@github.com:didix16/collections.git

# Fetch your patch-1 branches, the overkill way
git fetch --all

# Checkout your branch
git checkout patch-1

# Follow my previous instructions for squashing
git rebase -i origin/master

# Rework your commit message
git commit --amend

# Finally use the force
git push --force
```

didix16:
> I can totally do it for you if you don't feel comfortable jumping through the following hoops:
> 
> ```
> # Clone the repo
> git clone git@github.com:doctrine/collections.git
> 
> # Enter it
> cd collections
> 
> # Add your fork as a remote
> git remote add fork git@github.com:didix16/collections.git
> 
> # Fetch your patch-1 branch, the overkill way
> git fetch --all
> 
> # Checkout your branch
> git checkout patch-1
> 
> # Follow my previous instructions for squashing
> git rebase -i origin/master
> 
> # Rework your commit message
> git commit --amend
> 
> # Finally, use the force
> git push --force
> ```

Ok, thank you so much. I'll do it right now and see what happens :) :+1:

greg0ire (dismissed):

someniatko:
@greg0ire sorry, I afraid, those comments https://github.com/doctrine/collections/pull/221#issuecomment-554384144 https://github.com/doctrine/collections/pull/221#issuecomment-554377213 were not taken into account.

greg0ire:
Thanks!

lib/Doctrine/Common/Collections/ArrayCollection.php
SenseException: The tests are written in [PHPUnit](https://phpunit.readthedocs.io/en/8.4/) and can be found in `tests/Doctrine/Tests`. You will find that most of the time there is one test class that has the same name as the class you have changed, with the difference of a suffix called "Test".

In your case you have changed code in `lib/Doctrine/Common/Collections/ArrayCollection.php` and you'll find a file called `tests/Doctrine/Tests/Common/Collections/ArrayCollectionTest.php` that contains test methods. Because of the several types of collections which have to behave the same, it extends a `BaseArrayCollectionTest` class. Other test classes are extending `BaseCollectionTest`.

You'll have to create your new test method for your changes into the `BaseArrayCollectionTest` and `BaseCollectionTest`. You can take a look at `BaseCollectionTest::testMap()` to see how such a test can look like.

@Ocramius: Isn't `BaseArrayCollectionTest` supposed to extend `BaseCollectionTest`?

SenseException (changes requested):

VincentLanglet:
If we add keys to the map method, maybe the key should be the first param to add consistency with others methods.
See https://github.com/doctrine/collections/issues/251

greg0ire:
This seems stalled, I'm removing it from the 2.0.0 milestone.

PR #222, opened 2019-11-21
Fix of branch name for doctrine website build

SenseException:
This should fix the collections version infos and the build error `error: pathspec 'origin/1.6' did not match any file(s) known to git` on Doctrine website build.


alcaeus:
Thanks @SenseException!

PR #226, opened 2020-02-05
Remove stale branch alias

greg0ire:
This was useful when master was 1.6, but currently it is 2.0, so this
will only cause confusion. When releasing 2.0.0, we should create a
similar commit on the newly created 2.0.x branch, and bump that number
in the master branch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,10 +33,5 @@
         "psr-4": {
             "Doctrine\\Tests\\": "tests/Doctrine/Tests"
         }
-    },
-    "extra": {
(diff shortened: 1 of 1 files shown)


greg0ire:
I targeted 1.6.x since this should not affect the stability, and will create 2 upmerges PR for both 1.7.x and master subsequently.

greg0ire:
@alcaeus actually no upmerge PR for 1.7.x: it's fast-forward, which means now it's the very same commit.

PR #234, opened 2020-03-31
Add method find in ArrayCollection

Jeconias:
This enhancement gives the possibility of the found an element in ArrayCollection or remove any element of an easier and faster way.

A simple way to remove element using Doctrine (when all elements is found): 

```
$userEntity = $this->entityManager->getRepository(User::class)->findOneByUuid('123');

$tags = $userEntity->getTags(); // This method return a ArrayCollection of entity Tag

$tagFound = $tags->find(fn($element) => $element->getId() === '321');

$tags->removeElement($tagFound);
```

### DIFF
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -153,7 +153,7 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      *
      * @psalm-return array<TKey,T>
      */
-    public function toArray() : array;
+    public function toArray(): array;
 
     /**
      * Sets the internal iterator to the first element in the collection and returns this element.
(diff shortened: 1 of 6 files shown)


lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -224,6 +224,18 @@ public function exists(Closure $p) : bool;
...
+     * Return the value of first element found in array of elements that satisfies the $func.
+     *
+     * @param Closure $func The predicate used to find element.
+     *
+     * @return mixed|null The element if found or NULL if not.
+     *
+     * @psalm-param Closure(T=, TKey=):bool $func
+     * @psalm-return T|null
+     */
+    public function find(Closure $func);
```
alcaeus: This is a BC break, so we can't add this method in 1.x. We also don't have a strategy in place on how to add new methods to the collection interface without creating a hard BC break, which is why we've previously rejected all pull requests that expand the interface. Due to custom collection implementations (e.g. in ODM) we also can't just add the methods to some of these collection classes and "assume" they'll be there.

PR #238, opened 2020-06-11
Comparison BIT_AND

najved:
I was missing a "bit and" comparison for building criteria expression on bit mask fields.

e.g bit mask:

1: imported
2: canceled
4: deleted
8: ...

foo = 5 (1+4)

$expressionBuilder->bitAnd('foo', 1) // true
$expressionBuilder->bitAnd('foo', 5) // true
$expressionBuilder->bitAnd('foo', 2) // false

Implementation and tests included.

### DIFF
--- a/tests/Doctrine/Tests/Common/Collections/ClosureExpressionVisitorTest.php
+++ b/tests/Doctrine/Tests/Common/Collections/ClosureExpressionVisitorTest.php
@@ -198,7 +198,27 @@ class ClosureExpressionVisitorTest extends TestCase
         self::assertTrue($closure(new TestObject('4')));
         self::assertTrue($closure(new TestObject(5)));
         self::assertTrue($closure(new TestObject('5')));
+        self::assertTrue($closure(new TestObject(12)));
+        self::assertTrue($closure(new TestObject('12')));
         self::assertFalse($closure(new TestObject(2)));
+        self::assertFalse($closure(new TestObject('2')));
+        self::assertFalse($closure(new TestObject(8)));
+        self::assertFalse($closure(new TestObject('8')));
+        self::assertFalse($closure(new TestObject(11)));
+        self::assertFalse($closure(new TestObject('11')));
+
+        $closure = $this->visitor->walkComparison($this->builder->bitAnd('foo', 6));
+
+        self::assertTrue($closure(new TestObject(6)));
+        self::assertTrue($closure(new TestObject('6')));
+        self::assertTrue($closure(new TestObject(7)));
+        self::assertTrue($closure(new TestObject('7')));
+        self::assertFalse($closure(new TestObject(4)));
+        self::assertFalse($closure(new TestObject('4')));
+        self::assertFalse($closure(new TestObject(5)));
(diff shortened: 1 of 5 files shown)


tests/Doctrine/Tests/Common/Collections/ExpressionBuilderTest.php
```diff
@@ -141,4 +141,12 @@ public function testEndsWith() : void
         self::assertInstanceOf(Comparison::class, $expr);
         self::assertEquals(Comparison::ENDS_WITH, $expr->getOperator());
     }
+
+    public function testBitAnd() : void
+    {
+        $expr = $this->builder->bitAnd('b', 4);
+
+        self::assertInstanceOf(Comparison::class, $expr);
```
SenseException: The instance check shouldn't be needed in the test. It is a relict for past PHP versions when a return type couldn't be used.

tests/Doctrine/Tests/Common/Collections/ClosureExpressionVisitorTest.php
```diff
@@ -190,6 +190,17 @@ public function testWalkEndsWithComparison() : void
...
 
+    public function testWalkBitAndComparison() : void
+    {
+        $closure = $this->visitor->walkComparison($this->builder->bitAnd('foo', 4));
+
+        self::assertTrue($closure(new TestObject(4)));
+        self::assertTrue($closure(new TestObject('4')));
+        self::assertTrue($closure(new TestObject(5)));
+        self::assertTrue($closure(new TestObject('5')));
+        self::assertFalse($closure(new TestObject(2)));
```
SenseException, line 203: Please add a few more false cases like `8` or `11`. Be sure that there are also false-cases with a string-number like in your true-cases.

SenseException (changes requested):
Thank you for your PR. Can you please also elaborate which use-case you ran into that made you create this PR?

najved:
> Thank you for your PR. Can you please also elaborate which use-case you ran into that made you create this PR?

I have a collection of objects with a status field, which can hold different "combined values".

e.g.

```php
class Item {
  const IS_EXPORTED = 1;
  const IS_PAYMENT = 2;
  const IS_CANCELED = 4;

  $status = 3;
  $statusAsArray = [1, 2];
}

$expressionBuilder->bitAnd('status', Item::IS_PAYMENT + Item::IS_CANCELED);
$expressionBuilder->in('statusAsArray', [Item::IS_PAYMENT, Item::IS_CANCELED]) ;
```

"bitAnd" would be similar to "in" and "bitNotAnd" to "notIn".

While writing these lines, i'm not sure if this is just an "edge"-use-case in my legacy code or a good addition for this project :-)

SenseException (approved):
I don't know about the edge case, but I know that some doctrine/collections classes have a lot of methods which is why I don't know if it makes sense to introduce new ones or if the future of collections will need a different architecture. This needs more reviewers. @doctrine/doctrinecore

morozov:
Instead performing bitwise operations on a collection, why not have all bits as separate fields? This way, applied to a collection in the DB, the filter could use an index.

PR #241, opened 2020-06-21
[backport] Fix return type of Collection::key()

weirdan:
This PR proposes to backport the annotation fix by @enumag from master to 1.6.x. There's no functionality changed, it just fixes the docblock that was wrong in the first place (as `ArrayCollection::key()` does return `null` on empty collections).

PR #243, opened 2020-06-22
Use image url for a branch that actually exists

greg0ire:


greg0ire:
That's embarassing… should we do a release after this, so that the change bubbles up into master? There are some minor fixes that would improve everyone's life like psalm annotations from #241

malarzm (approved):

malarzm:
> should we do a release after this

Sure, why not :)

PR #249, opened 2020-07-25
Enable strict comparison for arrays of scalar values (1.6.x)

trickeyone:
This was originally created for #168, but I branched from v1.6.x and merged the previous changes into this PR to keep it in line with v1.6. It's the exact same changes from 168 just merged on top of v1.6.x

### DIFF
--- a/tests/Doctrine/Tests/Common/Collections/ClosureExpressionVisitorTest.php
+++ b/tests/Doctrine/Tests/Common/Collections/ClosureExpressionVisitorTest.php
@@ -143,6 +143,17 @@ class ClosureExpressionVisitorTest extends TestCase
         self::assertTrue($closure(new TestObject('04')));
     }
 
+    public function testWalkInComparisonObjects() : void
+    {
+        $closure = $this->visitor->walkComparison($this->builder->in('foo', [new TestObject(1), new TestObject(2), new TestObject(4)]));
+
+        self::assertTrue($closure(new TestObject(new TestObject(2))));
+        self::assertTrue($closure(new TestObject(new TestObject(1))));
+        self::assertFalse($closure(new TestObject(new TestObject(0))));
+        self::assertTrue($closure(new TestObject(new TestObject(4))));
+        self::assertFalse($closure(new TestObject(new TestObject('baz'))));
+    }
+
     public function testWalkNotInComparison() : void
     {
         $closure = $this->visitor->walkComparison($this->builder->notIn('foo', [1, 2, 3, '04']));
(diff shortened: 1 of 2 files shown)


greg0ire:
In light of #97 , I think the BC break here could be tolerated, since there is a bigger precedent (in #97)

greg0ire:
Tagging this as a bugfix since it fixes issues caused by the BC-break in #97

greg0ire:
@trickeyone I changed your commit message, does it accurately describe your changes?

tests/Doctrine/Tests/Common/Collections/ClosureExpressionVisitorTest.php
```diff
@@ -143,6 +143,17 @@ public function testWalkInComparison() : void
...
 
+    public function testWalkInComparisonObjects() : void
+    {
+        $closure = $this->visitor->walkComparison($this->builder->in('foo', [new TestObject(1), new TestObject(2)]));
+
+        self::assertTrue($closure(new TestObject(new TestObject(2))));
+        self::assertTrue($closure(new TestObject(new TestObject(1))));
+        self::assertFalse($closure(new TestObject(new TestObject(0))));
+        self::assertFalse($closure(new TestObject(4)));
+        self::assertFalse($closure(new TestObject('04')));
```
greg0ire: What is the point of the 2 last assertions? I might understand asserting an object is not confused with `4`, but why the need to compare it with `'04'`?

trickeyone:
@greg0ire Sorry, meant to respond yesterday. Yes, the commit message looks good.

tests/Doctrine/Tests/Common/Collections/ClosureExpressionVisitorTest.php
trickeyone: In all honesty, I can't completely recall. I'm fairly certain they came from the original tests that I duplicated, but I'm not sure why I didn't put them into a similar object structure as the previous assertions. I can correct that if you'd prefer.

greg0ire: I would prefer if everything made perfect sense for the test at hand. Not sure using a similar object structure will help though, I mean it's only interesting if you have 4 in the list where you look with the `IN` operator, and '04' as a thing you are looking for, isn't it?

trickeyone: Cool, no problem. Totally agree. I think I was wanting to add a negative assertion. I'll get it squared away today.

trickeyone: @greg0ire Sorry that took so long, other things came up today. Corrected the test and made it more sensible.

greg0ire (dismissed):

tests/Doctrine/Tests/Common/Collections/ClosureExpressionVisitorTest.php
greg0ire: Don't worry, I wasn't in a hurry :)

greg0ire:
Please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i origin/1.6.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

trickeyone:
@greg0ire commits squashed. Thanks!

greg0ire (approved):

ostrolucky (approved):

greg0ire:
Thanks @trickeyone !

Issue #231, opened 2020-03-13
[2.0] Callable instead of Closure


simPod:
Followup of https://github.com/doctrine/collections/issues/117

What about changing `Closure` types to `callable`? For example now I have a problem of passing `callable` into `Collection::map()` because static analysers complain about it.

greg0ire:
Closing as per https://github.com/doctrine/collections/pull/321#issuecomment-1227747872

Issue #190, opened 2019-04-10
[2.0] Use null instead of false on failure


enumag:
I noticed that some methods of `Collection` return `false` on failure. Specifically `first()`, `last()`, `current()`, `next()` and `indexOf()`. It would be better in my opinion to return `null` instead.

Since this is a BC break I'd like to propose this for doctrine/collections 2.0.

https://github.com/doctrine/collections/commit/c3564a69cf8e073f4031ce86a77fde44563f7f03
(Of course all `Collection` implementations and tests would need to be changed accordingly.)

ThomasLandauer:
Returning `null` would allow using PHP 8.0's [nullsafe operator](https://wiki.php.net/rfc/nullsafe_operator): `$foo->getBars()->first()?->getName();`

jeromegxj:
What is the status of this issue? I'm willing to implement it if needed.
Personally I'll be more than happy to have it, we may want to have methods like: `->firstOrNull()` to keep BC?
Let me know

derrabus:
I'm not aware that anyone is working on this issue, so feel free to do so.

jeromegxj:
A try has been given leading to a [discussion](https://github.com/doctrine/collections/pull/362#discussion_r1106339559)
The conclusion is that this is too much of a breaking change to be done even if it would go to the next major version.

mbolli:
Maybe a year later we could revisit this issue, or does the team still have the same opinion? 

This is one of my only gripes with Doctrine, always having to use 
`$obj->getRelation()->isEmpty() ? null : $obj->getRelation()->first()->getField();` 
instead of the much better readable
`$obj->getRelation()->first()?->getField()`

derrabus:
> Maybe a year later we could revisit this issue, or does the team still have the same opinion?

My opinion hasn't changed: https://github.com/doctrine/collections/pull/362#discussion_r1113998654

Issue #87, opened 2016-06-06
Add `each` method to the ArrayCollection


taylankasap:
Is there a reason for not adding it?

Ocramius:
BC compat, mostly. Collection interface is closed for modification.

Soullivaneuh:
> Collection interface is closed for modification.

Why not start implementation and add the method to the interface on next major?

Ocramius:
It would be interesting, but it would also have a high impact on the ecosystem, making upgrades very painful.
The collections API as-is needs some sort of redesign also to get rid of this kind of major BC break bumps for every new feature that is introduced.

Soullivaneuh:
> making upgrades very painful.

Why? Just one method to add for people implementing the interface directly, nothing to do for people using doctrine classes.

> The collections API as-is needs some sort of redesign also to get rid of this kind of major BC break bumps for every new feature that is introduced.

Which kind of redesign?

BTW, if you don't want to create a new major right now, you can implement the `each` method on each abstract implementing the interface and add the method to the interface but commented with a remember comment.

Then when you will release the next major, uncomment the method.

Not a very beautiful method, but this is the best way we found on sonata projects for now (cc @greg0ire @core23).

Ocramius:
> Which kind of redesign?

We looked a lot at things like https://github.com/morrisonlevi/Ardent - the collection API needs some sort of differentiation between dictionaries, arrays, stacks, linked lists, etc. The reason behind that being that the current logic is all crammed together into a single object that is just PHP arrays (again, sigh).

 > BTW, if you don't want to create a new major right now, you can implement the `each` method on each abstract implementing the interface and add the method to the interface but commented with a remember comment.

If we do that, people will start relying on a concrete implementation rather than an abstraction, and that would be more harmful to the ecosystem

greg0ire:
> If we do that, people will start relying on a concrete implementation rather than an abstraction, and that would be more harmful to the ecosystem

Alternatively, you can add a new interface, and merge it with the other on next major…

stof:
@Soullivaneuh all the Doctrine O*M projects are containing their own implementations of the collection API, and when using them, you have no guarantee about which implementation of the interface you will get. So adding a new method only in ArrayCollection here would make it unusable (if you use it, it will break when the ORM uses its own implementation).
And the different O*M projects are maintained by different teams, making it harder to enforce implicit contracts

Soullivaneuh:
[...]

PR #252, opened 2020-08-12
[2.0] Add new methods Collection::findFirst and Collection::reduce

VincentLanglet:
This is two methods I find useful for the next major.

Collection::find has a better performance than using Collection::filter then Collection::first.

Collection::reduce is similar to array_reduce.

I create this PR to open the discussion about these methods.

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -229,6 +229,18 @@ Applies the given function to each element in the collection and returns a new c
         return $value + 1;
     }); // [2, 3, 4]
 
+reduce
+------
+
+Applies iteratively the given function to each element in the collection, so as to reduce the collection to a single value.
+
+.. code-block:: php
+    $collection = new ArrayCollection([1, 2, 3]);
+
+    $reduce = $collection->reduce(function(int $accumulator, int $value): int {
+        return $accumulator + $value;
+    }, 0); // 6
+
 next
 ----
 
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -148,15 +148,15 @@ Tests for the existence of an element that satisfies the given predicate.
         return $value === 'first';
     }); // true
 
-findOne
--------
+findFirst
+---------
 
 Returns the first element of this collection that satisfies the given predicate.
 
 .. code-block:: php
     $collection = new Collection([1, 2, 3, 2, 1]);
 
-    $one = $collection->findOne(function(int $key, int $value): bool {
+    $one = $collection->findFirst(function(int $key, int $value): bool {
         return $value > 2 && $key > 1;
     }); // 3
 
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -148,6 +148,18 @@ Tests for the existence of an element that satisfies the given predicate.
         return $value === 'first';
     }); // true
 
+findOne
+-------
+
+Returns the first element of this collection that satisfies the given predicate.
+
+.. code-block:: php
+    $collection = new Collection([1, 2, 3, 2, 1]);
+
+    $one = $collection->findOne(function(int $key, int $value): bool {
+        return $value > 2 && $key > 1;
+    }); // 3
+
 filter
 ------
 
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -213,6 +213,19 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
     public function exists(Closure $p): bool;
 
     /**
+     * Returns the first element of this collection that satisfies the predicate p.
+     *
+     * @param Closure $p The predicate.
+     *
+     * @return mixed The first element respecting the predicate,
+     *               null if no element respects the predicate.
+     *
+     * @psalm-param Closure(TKey=, T=):bool $p
+     * @psalm-return T|null
+     */
+    public function findOne(Closure $p);
+
+    /**
      * Returns all the elements of this collection that satisfy the predicate p.
      * The order of the elements is preserved.
      *
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -262,6 +262,22 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
     public function map(Closure $func): self;
 
     /**
+     * Applies iteratively the given function to each element in the collection,
+     * so as to reduce the collection to a single value.
+     *
+     * @param mixed $initial
+     *
+     * @return mixed
+     *
+     * @psalm-template TReturn
+     * @psalm-template TInitial
+     * @psalm-param Closure(TReturn|TInitial, T):TReturn $func
+     * @psalm-param TInitial|null $initial
+     * @psalm-return TReturn|TInitial|null
+     */
+    public function reduce(Closure $func, $initial = null);
+
+    /**
      * Partitions this collection in two collections according to a predicate.
      * Keys are preserved in the resulting collections.
      *
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -271,7 +271,7 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      *
      * @psalm-template TReturn
      * @psalm-template TInitial
-     * @psalm-param Closure(TReturn|TInitial, T):TReturn $func
+     * @psalm-param Closure(TReturn|TInitial|null, T):(TInitial|TReturn) $func
      * @psalm-param TInitial|null $initial
      * @psalm-return TReturn|TInitial|null
      */
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -223,7 +223,7 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * @psalm-param Closure(TKey=, T=):bool $p
      * @psalm-return T|null
      */
-    public function findOne(Closure $p);
+    public function findFirst(Closure $p);
 
     /**
      * Returns all the elements of this collection that satisfy the predicate p.
--- a/tests/Doctrine/Tests/Common/Collections/BaseArrayCollectionTest.php
+++ b/tests/Doctrine/Tests/Common/Collections/BaseArrayCollectionTest.php
@@ -278,22 +278,22 @@ abstract class BaseArrayCollectionTest extends TestCase
         }), 'Element not exists');
     }
 
-    public function testFindOne(): void
+    public function testFindFirst(): void
     {
         $elements   = [1, 'A' => 'a', 2, 'null' => null, 3, 'A2' => 'a', 'zero' => 0];
         $collection = $this->buildCollection($elements);
 
-        self::assertSame('a', $collection->findOne(static function ($key, $element) {
+        self::assertSame('a', $collection->findFirst(static function ($key, $element) {
             return $key === 'A' && $element === 'a';
         }), 'Element exists');
     }
 
-    public function testFindOneNotFound(): void
+    public function testFindFirstNotFound(): void
     {
         $elements   = [1, 'A' => 'a', 2, 'null' => null, 3, 'A2' => 'a', 'zero' => 0];
         $collection = $this->buildCollection($elements);
 
-        self::assertNull($collection->findOne(static function ($key, $element) {
+        self::assertNull($collection->findFirst(static function ($key, $element) {
             return $key === 'non-existent' && $element === 'non-existent';
         }), 'Element does not exists');
     }
--- a/tests/Doctrine/Tests/Common/Collections/BaseArrayCollectionTest.php
+++ b/tests/Doctrine/Tests/Common/Collections/BaseArrayCollectionTest.php
@@ -278,6 +278,26 @@ abstract class BaseArrayCollectionTest extends TestCase
         }), 'Element not exists');
     }
 
+    public function testFindOne(): void
+    {
+        $elements   = [1, 'A' => 'a', 2, 'null' => null, 3, 'A2' => 'a', 'zero' => 0];
+        $collection = $this->buildCollection($elements);
+
+        self::assertSame('a', $collection->findOne(static function ($key, $element) {
+            return $key === 'A' && $element === 'a';
+        }), 'Element exists');
+    }
+
+    public function testFindOneNotFound(): void
+    {
+        $elements   = [1, 'A' => 'a', 2, 'null' => null, 3, 'A2' => 'a', 'zero' => 0];
+        $collection = $this->buildCollection($elements);
+
+        self::assertNull($collection->findOne(static function ($key, $element) {
+            return $key === 'non-existent' && $element === 'non-existent';
+        }), 'Element does not exists');
+    }
+
     public function testIndexOf(): void
     {
         $elements   = [1, 'A' => 'a', 2, 'null' => null, 3, 'A2' => 'a', 'zero' => 0];
--- a/tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
+++ b/tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
@@ -76,6 +76,19 @@ abstract class BaseCollectionTest extends TestCase
         self::assertEquals([2, 4], $res->toArray());
     }
 
+    public function testReduce(): void
+    {
+        $this->collection->add(1);
+        $this->collection->add(2);
+        $this->collection->add(3);
+        $this->collection->add(4);
+
+        $res = $this->collection->reduce(static function ($sum, $e) {
+            return $sum + $e;
+        });
+        self::assertSame(10, $res);
+    }
+
     public function testFilter(): void
     {
         $this->collection->add(1);
--- a/tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
+++ b/tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
@@ -46,6 +46,26 @@ abstract class BaseCollectionTest extends TestCase
         self::assertFalse($exists);
     }
 
+    public function testFindOne(): void
+    {
+        $this->collection->add('one');
+        $this->collection->add('two');
+        $one = $this->collection->findOne(static function ($k, $e) {
+            return $e === 'one';
+        });
+        self::assertSame('one', $one);
+    }
+
+    public function testFindOneNotFound(): void
+    {
+        $this->collection->add('one');
+        $this->collection->add('two');
+        $other = $this->collection->findOne(static function ($k, $e) {
+            return $e === 'other';
+        });
+        self::assertNull($other);
+    }
+
     public function testMap(): void
     {
         $this->collection->add(1);
--- a/tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
+++ b/tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
@@ -46,21 +46,21 @@ abstract class BaseCollectionTest extends TestCase
         self::assertFalse($exists);
     }
 
-    public function testFindOne(): void
+    public function testFindFirst(): void
     {
         $this->collection->add('one');
         $this->collection->add('two');
-        $one = $this->collection->findOne(static function ($k, $e) {
+        $one = $this->collection->findFirst(static function ($k, $e) {
             return $e === 'one';
         });
         self::assertSame('one', $one);
     }
 
-    public function testFindOneNotFound(): void
+    public function testFindFirstNotFound(): void
     {
         $this->collection->add('one');
         $this->collection->add('two');
-        $other = $this->collection->findOne(static function ($k, $e) {
+        $other = $this->collection->findFirst(static function ($k, $e) {
             return $e === 'other';
         });
         self::assertNull($other);
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -204,6 +204,13 @@ abstract class AbstractLazyCollection implements Collection
         return $this->collection->exists($p);
     }
 
+    public function findOne(Closure $func)
+    {
+        $this->initialize();
+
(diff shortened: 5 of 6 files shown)


lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -212,6 +212,19 @@ public function next();
      */
     public function exists(Closure $p) : bool;
 
+    /**
+     * Returns the first elements of this collection that satisfy the predicate p.
```
greg0ire:
```suggestion
     * Returns the first element of this collection that satisfies the predicate p.
```

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -212,6 +212,19 @@ public function next();
...
+     *
+     * @param Closure $p The predicate.
+     *
+     * @return mixed The first elements respecting the predicate,
```
greg0ire:
```suggestion
     * @return mixed The first element respecting the predicate,
```

tests/Doctrine/Tests/Common/Collections/BaseArrayCollectionTest.php
```diff
@@ -269,6 +269,20 @@ public function testExists() : void
...
+        $elements   = [1, 'A' => 'a', 2, 'null' => null, 3, 'A2' => 'a', 'zero' => 0];
+        $collection = $this->buildCollection($elements);
+
+        self::assertSame('a', $collection->find(static function ($key, $element) {
+            return $key === 'A' && $element === 'a';
+        }), 'Element exists');
+
+        self::assertFalse($collection->find(static function ($key, $element) {
+            return $key === 'non-existent' && $element === 'non-existent';
+        }), 'Element not exists');
```
greg0ire:
```suggestion
        }), 'Element does not exist');
```

greg0ire (commented):
You should probably add docs at https://github.com/doctrine/collections/blob/1.6.x/docs/en/index.rst

SenseException:
> You should probably add docs at https://github.com/doctrine/collections/blob/1.6.x/docs/en/index.rst

Probably the 2.0 docs if this is for the next major release.

VincentLanglet:
I added the documentation.

tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
```diff
@@ -55,6 +69,16 @@ public function testMap() : void
         self::assertEquals([2, 4], $res->toArray());
     }
 
+    public function testReduce() : void
+    {
+        $this->collection->add(1);
+        $this->collection->add(2);
```
SenseException, line 82: Can you expand this with more elements please? It should make the test more expressive.

tests/Doctrine/Tests/Common/Collections/BaseArrayCollectionTest.php
```diff
@@ -269,6 +269,20 @@ public function testExists() : void
         }), 'Element not exists');
     }
 
+    public function testFind() : void
```
SenseException: Please split up into separate tests for "found" and "not found".

tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
```diff
@@ -45,6 +45,20 @@ public function testExists() : void
         self::assertFalse($exists);
     }
 
+    public function testFind() : void
```
SenseException: Please split up into separate tests for "found" and "not found".

docs/en/index.rst
```diff
@@ -148,6 +148,18 @@ Tests for the existence of an element that satisfies the given predicate.
         return $value === 'first';
     }); // true
 
+find
+----
+
+Returns the first element of this collection that satisfies the given predicate.
+
+.. code-block:: php
+    $collection = new Collection(['first', 'second', 'third']);
```
SenseException: Can you please use an example where the element, that is searched for, is twice or more in the collection? It would emphasize the chosen behavior.

SenseException (changes requested):

tests/Doctrine/Tests/Common/Collections/BaseArrayCollectionTest.php
VincentLanglet: All the others tests are not split.
- `testRemove` try to remove 5 elements
- `testEmpty` is called on empty and not empty
- `testExist` is almost a copy paste of my test

And so on...

I feel like it would be inconsistent to split this one.

docs/en/index.rst
VincentLanglet: Sure, done

tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
VincentLanglet: Sure, done.

tests/Doctrine/Tests/Common/Collections/BaseArrayCollectionTest.php
SenseException: Should have been, but fixing those is out of scope for your PR.

VincentLanglet: Done

docs/en/index.rst
```diff
@@ -148,6 +148,18 @@ Tests for the existence of an element that satisfies the given predicate.
         return $value === 'first';
     }); // true
 
+find
```
greg0ire: Maybe we should name it `findOne` (as opposed to `findAll()`?)

docs/en/index.rst
```diff
@@ -148,6 +148,18 @@ Tests for the existence of an element that satisfies the given predicate.
...
+----
+
+Returns the first element of this collection that satisfies the given predicate.
+
+.. code-block:: php
+    $collection = new Collection([1, 2, 3, 2, 1]);
+
+    $one = $collection->find(function($key, $value) {
+        return $value === 1;
+    }); // 1
```
greg0ire: It would be great to have an example which uses `$key`

docs/en/index.rst
```diff
@@ -148,6 +148,18 @@ Tests for the existence of an element that satisfies the given predicate.
...
 
+find
+----
+
+Returns the first element of this collection that satisfies the given predicate.
+
+.. code-block:: php
+    $collection = new Collection([1, 2, 3, 2, 1]);
+
+    $one = $collection->find(function($key, $value) {
```
greg0ire: Let's use type declarations?

docs/en/index.rst
```diff
@@ -216,6 +228,18 @@ Applies the given function to each element in the collection and returns a new c
...
 
+reduce
+------
+
+Applies iteratively the given function to each element in the collection, so as to reduce the collection to a single value.
+
+.. code-block:: php
+    $collection = new ArrayCollection([1, 2, 3]);
+
+    $reduce = $collection->reduce(function($accumulator, $value) {
```
greg0ire: Type declarations here too please?

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -212,6 +212,19 @@ public function next();
...
 
+    /**
+     * Returns the first element of this collection that satisfies the predicate p.
+     *
+     * @param Closure $p The predicate.
+     *
+     * @return mixed The first element respecting the predicate,
+     *               FALSE if no element respects the predicate.
+     *
+     * @psalm-param Closure(TKey=, T=):bool $p
```
greg0ire, line 223: Where can I learn more about this `=` notation? It's the first time I see it I think

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -212,6 +212,19 @@ public function next();
      */
     public function exists(Closure $p) : bool;
 
+    /**
+     * Returns the first element of this collection that satisfies the predicate p.
+     *
+     * @param Closure $p The predicate.
+     *
+     * @return mixed The first element respecting the predicate,
```
greg0ire, line 220: What if the first element respecting the predicate is literally `false` itself? Wouldn't it be better to throw an exception that recommends to call whatever method allows to check that there will be a result? (`exists`?)

lib/Doctrine/Common/Collections/Collection.php
VincentLanglet: I don't know. I don't use psalm.
Every other phpdoc use this notation so I did the same.

greg0ire: Apparently it makes them optional: 744afc199626c1638113da4cd4bf2222c91dcd13

lib/Doctrine/Common/Collections/Collection.php
VincentLanglet: I don't like to require an `exists` call, it add complexity to the code and it lower performance (the search is made twice).

Collection is not made to store boolean.
All the method `current/next/first/last/...` return `T|false`.

But maybe, It's better to return `null`.
The method `get` is returning `T|null`.

VincentLanglet: I changed for `null` if not found.

docs/en/index.rst
VincentLanglet: Done

lib/Doctrine/Common/Collections/Collection.php
greg0ire: >  All the method current/next/first/last/... return T|false.

Ok, if this is an issue, it's an issue of doctrine/collections, not this PR

greg0ire (dismissed):

lib/Doctrine/Common/Collections/Collection.php
someniatko: @greg0ire https://psalm.dev/docs/annotating_code/type_syntax/callable_types/

greg0ire (dismissed):

SenseException (dismissed):

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -212,6 +212,19 @@ public function next();
...
+     *               null if no element respects the predicate.
+     *
+     * @psalm-param Closure(TKey=, T=):bool $p
+     * @psalm-return T|null
+     */
+    public function findOne(Closure $p);
```
ostrolucky: I'm wondering how was new `findOne` method name chosen. Name doesn't indicate it will return first matching result. For all I know, it can return last matching result. In fact, it's doing same thing as `first()`, except it skips non-matching results. I'm especially weirded out by this choice, because even laravel/collections uses `first(callable)` for this.

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -248,6 +261,21 @@ public function forAll(Closure $p) : bool;
...
+     * so as to reduce the collection to a single value.
+     *
+     * @param mixed $initial
+     *
+     * @return mixed
+     *
+     * @psalm-template U
+     * @psalm-param Closure(U=,T=):U $func
+     * @psalm-param U $initial
+     * @psalm-return U
```
ostrolucky: This is not accurate. If Collection is empty, array_reduce will return `null`, no matter what initial value you specify.

Also, no matter what initial value you specify, closure can return some completely different data type than initial value.

This is more like it https://github.com/vimeo/psalm/issues/846

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -212,6 +212,19 @@ public function next();
...
+    /**
+     * Returns the first element of this collection that satisfies the predicate p.
+     *
+     * @param Closure $p The predicate.
+     *
+     * @return mixed The first element respecting the predicate,
+     *               null if no element respects the predicate.
+     *
+     * @psalm-param Closure(TKey=, T=):bool $p
+     * @psalm-return T|null
```
ostrolucky, line 224: `findOne` returning null when result is not found makes it:
- Inconsistent with what `last()` and `first()` methods return. I think we should first solve this inconsistency before merging this PR
- Impossible to differentiate between missing result and collection containing `null` value

ostrolucky (changes requested):

lib/Doctrine/Common/Collections/Collection.php
VincentLanglet: I previously return `false` but got a review from @greg0ire https://github.com/doctrine/collections/pull/252#discussion_r475088024

I can return `false` for consistency, if you prefer. You just need to decide.

By the way, `last` and `first` returns false if not found ; but `get` return null if not found.

lib/Doctrine/Common/Collections/Collection.php
VincentLanglet: I previously named this `find`. Then @greg0ire asked for `findOne` https://github.com/doctrine/collections/pull/252#discussion_r475087688.

I don't think as weird the term `find`. It seems familiar to me.
This is similar to JS: https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/find
Or to doctrine/ORM `findOne` method.

IMHO, the fact that this is the *first* matching result should be something relevant.
I introduced this method because
- You may need to know that the element exists AND you need to use it.
- Using both `filter` and `first` are under-performant for this case.

lib/Doctrine/Common/Collections/Collection.php
ostrolucky: Yeah I know we need to decide first. I'm fine with returning null, but we should first make existing code consistent so it returns null, otherwise we create more mess by merging this PR. I do have enough experience in software engineering to know better than to believe in "we will clean it up... later"

lib/Doctrine/Common/Collections/Collection.php
ostrolucky: JS doesn't already have `first` function though. ORM does have `find` function, but it has completely different signature than this one. In fact, without knowing what these functions do beforehand, I wouldn't be able to tell the difference between `Collection::findOne` and `ObjectRepository::find` based on name, or I would think `ObjectRepository::find` returns more results than `Collection::findOne`, then be confused why one expects ID and another one closure.

@greg0ire ok to switch back to `first`?

VincentLanglet: > @greg0ire ok to switch back to `first`?

`first` already exist, it
```
Sets the internal iterator to the first element in the collection and returns this element.
```

The method doesn't modify the internal iterator, and should not.
I don't think it's a good idea to modify the signature from `first()` to `first(?callable)`.

A specific function seems better.

> ORM does have find function, but it has completely different signature than this one. In fact, without knowing what these functions do beforehand, I wouldn't be able to tell the difference between Collection::findOne and ObjectRepository::find based on name, or I would think ObjectRepository::find returns more results than Collection::findOne, then be confused why one expects ID and another one closure.

I don't know why you compare to `find`, you should compare to `findOne` instead.
Both `Collection::findOne` and `ObjectRepository::findOne` will return 1 value.

lib/Doctrine/Common/Collections/Collection.php
VincentLanglet: I can see some logic in the current situation.

When we want to access to a value (like a `get`), `null` is return when not found.
When we want to perform an action 'like a `set`), `false` is return when a failure occurs.

`first` sets the index to the first element. If there is no element, it **cannot set**. Failure => `false`.
Same idea for `last`, `current`, `next`.

`key` gets the key of the element at the current iterator position. If there is no element, it **cannot get**. Nothing =>`null`.
Same idea for `get`. It would be the same idea for `findOne`.

The only remaining confusing method would be `remove` then.
It removes the element at a specific `key`, and return `null` if the element was already removed. By following the same logic, it should return `false` instead. (By the way, `removeElement` return `false` for the similar situation).
But [that's not the only issue I have with the `remove` method](https://github.com/doctrine/collections/issues/87#issuecomment-673672873): For consistency, I would prefer `remove` and `removeKey` instead of `removeElement` and `remove`.

lib/Doctrine/Common/Collections/Collection.php
ostrolucky: > I don't know why you compare to find, you should compare to findOne instead.

`ObjectRepository::findOne` doesn't exist either. Are you thinking of `findOneBy`?

> The method doesn't modify the internal iterator, and should not.

This is good argument, but TBH I don't see the utility of functions like current(), next() and key() or other ways to play with internal pointer. I never saw a use for those, it was always foreach. We might need to think of removing them. Then nobody would have issues related to internal pointer, as it wouldn't be exposed.

> IMHO, the fact that this is the first matching result should be something relevant.

WDYM? This is why I suggested to use `first`. I don't complain about `One` part, I'm on board with that. Otherwise `findOne` name doesn't tell it returns first matching result. Yes, it's consistent with current implicit Persistence API, doesn't mean we can't create better, more explicit API going forward. Perhaps `findFirst()`?

VincentLanglet: > > I don't know why you compare to find, you should compare to findOne instead.
> 
> `ObjectRepository::findOne` doesn't exist either. Are you thinking of `findOneBy`?

Yes indeed ! I was thinking of `findOneBy`, sorry for my mistake.

> > The method doesn't modify the internal iterator, and should not.
> 
> This is good argument, but TBH I don't see the utility of functions like current(), next() and key() or other ways to play with internal pointer. I never saw a use for those, it was always foreach. We might need to think of removing them. Then nobody would have issues related to internal pointer, as it wouldn't be exposed.

I never played with iterator either.

> > IMHO, the fact that this is the first matching result should be something relevant.
> 
> WDYM?

I was kinda asking myself if the fact that it was the first matching result was something important.

For instance if I had something as 
```
new ArrayCollection(
    ['id' => 1, 'ready' => true],
    ['id' => 2, 'ready' => false],
    ['id' => 3, 'ready' => false],
    ['id' => 4, 'ready' => true],
    ['id' => 5, 'ready' => true],
)
```
And I look for a slot which is ready, I won't care if it's the id `1`, `4` or `5`.
I also don't care if sometimes this configuration gives me `1` and sometimes it gives me `5`. I just want one.

If the fact that it's the `first` result is not relevant, `findOne` seems ok.
And then maybe the doc should be updated to
```
Returns one element of this collection that satisfies the given predicate.
```

If the fact the it's the `first` result is important, `findFirst` could be indeed better.

VincentLanglet:
@ostrolucky Sorry for the delay ; I made the changes you asked :)

lib/Doctrine/Common/Collections/Collection.php
VincentLanglet: Php does have a `array_search` method ; maybe `searchValue` is better ?

lib/Doctrine/Common/Collections/Collection.php
VincentLanglet: Done

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -248,6 +261,22 @@ public function forAll(Closure $p) : bool;
...
+     * Applies iteratively the given function to each element in the collection,
+     * so as to reduce the collection to a single value.
+     *
+     * @param mixed $initial
+     *
+     * @return mixed
+     *
+     * @psalm-template TReturn
+     * @psalm-template TInitial
+     * @psalm-param Closure():TReturn $func
```
ostrolucky: Parameters are required to specify. Please make sure whole definition works in psalm.dev https://psalm.dev/r/61a6394fe6

ostrolucky (changes requested):

ostrolucky:
Still not it https://psalm.dev/r/b66b6e6a60

VincentLanglet:
> Still not it https://psalm.dev/r/b66b6e6a60

I think your test is wrong since you're passing string to array_reduce. @ostrolucky 

See https://psalm.dev/r/6ca84608e3

ostrolucky (dismissed):

VincentLanglet:
Is there anything missing to be merged ?

ostrolucky (approved):

ostrolucky:
I don't think so. I am not sure who is responsible for doctrine/collections though.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @VincentLanglet !

VincentLanglet:
> Thanks @VincentLanglet !

You're welcome.

By the way, I saw multiple times that creating a release 2.0 seems impossible since the library will have difficulties to migrate from version 1 and 2. But when I looked at the 2.0 branches, the changes are currently:
- Some method are added to the interface.
- Some return type hint are added.
Which means that if you're compatible with the 2.0 version, you're still compatible with the v1.

What about focusing the feature of the v2 on this point ? (Being compatible with the v2 should allow to be compatible with the v1 easily). For instance: 
- https://github.com/doctrine/collections/issues/231 ; can be easily done since a Closure is a callable, if the method supports the callable, it still supports the closure and work with the v1.
- https://github.com/doctrine/collections/issues/190 ; can not be done since if you return null, you're not compatible with the v1. 

With this strategy, a reachable 2.0 milestone could be created and released ; Wdyt ?

ostrolucky:
If 2.0 is created now, it will mean we have to create 3.0 for new features/BC breaks, so that's unlikely. It's not worth creating 2.0 at this point IMHO, there are not that many changes.

VincentLanglet:
> If 2.0 is created now, it will mean we have to create 3.0 for new features/BC breaks, so that's unlikely. It's not worth creating 2.0 at this point IMHO, there are not that many changes.

I didn't ask for creating 2.0 now. But for a reachable milestone.
Almost 0 issue are flagged with the milestone https://github.com/doctrine/collections/issues?q=is%3Aissue+milestone%3A2.0.0+is%3Aopen and the only one will break the support of 1.x.

The last major version was in 2013, I'm sure they are some things which can be improved and allow librairies to support both `^1 || ^2`.

Brammm:
Derp, I was just about to create a PR for a similar findFirst method. Is there any reason this is going in 2.0 and not 1.x?

VincentLanglet:
> Derp, I was just about to create a PR for a similar findFirst method. Is there any reason this is going in 2.0 and not 1.x?

Adding a method to an interface is a BC-break.

Brammm:
Hadn't considered that. Thanks for explaining.

fnagel:
@VincentLanglet Both methods are not part of the UPGRADE.md file or am I missing something?

PR #253, opened 2020-09-02
Add method to create a new collection with fixed index

skuhnow:
Add a method to create a new collection with rebuilded keys.

If you use a method that unsets an element in the collection, the keys are not continuous anymore. This can create problems if you serialize an ArrayCollection as an JSON. For JavaScript this would be an object and not an array anymore.

### DIFF
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -438,4 +438,12 @@ class ArrayCollection implements Collection, Selectable
 
         return $this->createFrom($filtered);
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


malarzm:
I'd argue this does not belong to the `ArrayCollection` but rather there should be a list implementation of a `Collection` that will take care of keeping consecutive indices. FWIW I did it like this https://github.com/malarzm/collections/blob/master/lib/Malarzm/Collections/ListArray.php. In context of ODM/ORM, ODM allows one to specify what collection class should be used (https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.1/reference/custom-collections.html) but this is not available in ORM yet.

PR #254, opened 2020-09-09
Update index.rst

ThomasLandauer:
* Renaming "count" to "number", since "count" was ambiguous (meaning: how many?)
* Explaining the callback function of `filter()` better

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -12,8 +12,8 @@ arrays of data. Here is an example using the simple
 
     $collection = new ArrayCollection([1, 2, 3]);
 
-    $filteredCollection = $collection->filter(function($count) {
-        return $count > 1;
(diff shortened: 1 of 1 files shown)


SenseException (changes requested):
Thank you for the contribution. I agree that `$count` doesn't seem to add to readability, but `$number` neither. The argument of the anonymous function gets every element of the collection injected. What do you think about `$element`, `$item`, `$collectionItem` or `$collectionElement`?

ThomasLandauer:
Sure! :-)
What I meant: `count` is *ambiguous* - I was first thinking what's injected is the *number of elements* (not the element itself)

SenseException (approved):

SenseException:
Thank you @ThomasLandauer

PR #258, opened 2020-11-29
Maintenance work

greg0ire:


greg0ire:
To see the result, take a look at https://github.com/greg0ire/collections/pull/1

phpstan.neon.dist
```diff
@@ -0,0 +1,17 @@
...
+            message: '~Unsafe usage of new static\(\)\.~'
+            paths:
+                - 'lib/Doctrine/Common/Collections/ArrayCollection.php'
+                - 'lib/Doctrine/Common/Collections/Criteria.php'
+        -
+            message: '~Array \(array\<TKey, T\>\) does not accept key int\.~'
+            path: 'lib/Doctrine/Common/Collections/ArrayCollection.php'
+        -
+            message: '~Method Doctrine\\Common\\Collections\\ArrayCollection\:\:map\(\) should return Doctrine\\Common\\Collections\\ArrayCollection\<TKey, U\> but returns Doctrine\\Common\\Collections\\ArrayCollection\<TKey, T\>\.~'
+            path: 'lib/Doctrine/Common/Collections/ArrayCollection.php'
```
greg0ire: @weirdan @someniatko if maybe you could have a look at these errors I ignored, it would be great.

phpstan.neon.dist
```diff
@@ -0,0 +1,17 @@
+parameters:
+    level: 3
+    paths:
+        - lib
+    ignoreErrors:
+        # Making classes final as suggested would be a BC-break
+        -
+            message: '~Unsafe usage of new static\(\)\.~'
```
weirdan, line 8: This is a valid issue. Do you know if PHPStan supports anything like [`@psalm-consistent-constructor`](https://psalm.dev/docs/running_psalm/issues/UnsafeInstantiation/)?

phpstan.neon.dist
```diff
@@ -0,0 +1,17 @@
...
+            paths:
+                - 'lib/Doctrine/Common/Collections/ArrayCollection.php'
+                - 'lib/Doctrine/Common/Collections/Criteria.php'
+        -
+            message: '~Array \(array\<TKey, T\>\) does not accept key int\.~'
```
weirdan, line 13: This is also suppressed for Psalm, so there doesn't seem to be a better way.

phpstan.neon.dist
```diff
@@ -0,0 +1,17 @@
...
+            path: 'lib/Doctrine/Common/Collections/ArrayCollection.php'
+        -
+            message: '~Method Doctrine\\Common\\Collections\\ArrayCollection\:\:map\(\) should return Doctrine\\Common\\Collections\\ArrayCollection\<TKey, U\> but returns Doctrine\\Common\\Collections\\ArrayCollection\<TKey, T\>\.~'
```
weirdan: This one should not be suppressed, but rather fixed in `createFrom()` docblock. The following needs to be added/changed:
```
     * @psalm-template K of array-key
     * @psalm-template V
     * @phpstan-template K
     *
     * @psalm-param array<K,V> $elements
     * @psalm-return static<K,V>
```

phpstan.neon.dist
weirdan: Generally speaking, that `add()` method is broken looking from the type safety perspective. E.g. when you have `Collection<string,string>` and you call `add()` on it you're in fact mutating it into `Collection<string|int,string>`, invalidating assumption about types being constant.

phpstan.neon.dist
greg0ire: Good question, I asked it at https://symfony-devs.slack.com/archives/C8SFXTD2M/p1606661194462100

greg0ire: I have tried using `@psalm-consistent-constructor` for Psalm, it only worked for `ArrayCollection`, not for `Criteria`…

phpstan.neon.dist
greg0ire: Indeed, that's not great…

phpstan.neon.dist
weirdan: Works for me in both cases:
```console
> vendor/bin/psalm --show-info=true --no-cache                                                                                                                                 
Scanning files...
Analyzing files...

░░░░II

INFO: UnsafeInstantiation - lib/Doctrine/Common/Collections/ArrayCollection.php:100:16 - Cannot safely instantiate class Doctrine\Common\Collections\ArrayCollection with "new static" as its constructor might change in child classes (see https://psalm.dev/229)
        return new static($elements);


INFO: UnsafeInstantiation - lib/Doctrine/Common/Collections/Criteria.php:42:16 - Cannot safely instantiate class Doctrine\Common\Collections\Criteria with "new static" as its constructor might change in child classes (see https://psalm.dev/229)
        return new static();


No errors found!
2 other issues found.

Checks took 1.45 seconds and used 45.516MB of memory
Psalm was able to infer types for 99.2840% of the codebase

> git stash pop                                                                                                                                                                
On branch maintenance-work
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
   modified:   lib/Doctrine/Common/Collections/ArrayCollection.php
   modified:   lib/Doctrine/Common/Collections/Criteria.php

no changes added to commit (use "git add" and/or "git commit -a")
Dropped refs/stash@{0} (33260b3f4a197fcda746946e543f945b727f3659)

> git diff                                                                                                                                                                     
diff --git a/lib/Doctrine/Common/Collections/ArrayCollection.php b/lib/Doctrine/Common/Collections/ArrayCollection.php
index 963e3d1..6a16d33 100644
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -39,6 +39,7 @@
  * @psalm-template T
+ * @psalm-consistent-constructor
  */
 class ArrayCollection implements Collection, Selectable
 {
diff --git a/lib/Doctrine/Common/Collections/Criteria.php b/lib/Doctrine/Common/Collections/Criteria.php
index 8a5324a..d0aa960 100644
--- a/lib/Doctrine/Common/Collections/Criteria.php
+++ b/lib/Doctrine/Common/Collections/Criteria.php
@@ -10,6 +10,7 @@
 
 /**
  * Criteria for filtering Selectable collections.
+ * @psalm-consistent-constructor
  */
 class Criteria
 {

> vendor/bin/psalm --show-info=true --no-cache                                                                                                                                 
Scanning files...
Analyzing files...

░░░░░░
No errors found!

Checks took 1.45 seconds and used 45.516MB of memory
Psalm was able to infer types for 99.2840% of the codebase
```

greg0ire: Ah, the `--no-cache` was key. Thanks!

phpstan.neon.dist
someniatko: Oh, thank you for invitation, it was a bit of unexpected! I guess weirdan is right, as long as `createFrom()` function is a static one, it should be "templated" by its own types instead of TKey and T of the class.

greg0ire: Yup, done! Thanks for chiming in!

phpstan.neon.dist
greg0ire: Another way to fix that without resorting to annotations would be to use an interface: https://phpstan.org/r/2857e582-2ede-41e2-82b6-ad4d6cda1227

I probably shouldn't do it in a patch release but it might be a good thing to do on the 1.7.x

phpstan.neon.dist
weirdan: > as long as createFrom() function is a static one, it should be "templated" by its own types instead of TKey and T of the class.

It's not static, but it creates a new, unrelated collection instance. Thus it should not be constrained by the types of the instance it was called on.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,93 @@
...
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer update --no-interaction --prefer-dist"
+
+      - name: "Run PHPUnit"
+        run: "vendor/bin/phpunit --coverage-clover=coverage.xml"
```
alcaeus, line 65: Unless we have people looking at coverage, we might as well not bother generating reports.

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,61 @@
...
+      matrix:
+        php-version:
+          - "7.4"
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Psalm
+        uses: docker://vimeo/psalm-github-actions:4.2.1
```
alcaeus: This should invoke psalm as defined in composer.json instead of separately tracking the dependency. This is bound to get out of sync at some point in time (for example when adding dependabot).

alcaeus (commented):
Just took a quick glance, will review in more detail soon.

phpstan.neon.dist
someniatko: Oh, I was about to write "behaves as a static function", but had actually written not what i wanted 😌

.github/workflows/static-analysis.yml
greg0ire: I thought we were going to have a discussion about that at https://github.com/doctrine/.github , so that make we make that change on all repositories.

.github/workflows/continuous-integration.yml
greg0ire: We mostly use PCOV now, so it's not a big deal (it does not slow down), and CodeCov will comment on PRs, telling people when their code is not covered. On https://github.com/doctrine/common/pull/910, that helped me notice that tests were not run on PHP 8 :stuck_out_tongue:  . So while I'm not sure people will look at coverage, IMO it can be useful.

.github/workflows/static-analysis.yml
alcaeus: Sounds good, I'm ok with keeping this as-is at the moment.

greg0ire: Or let's wait for the outcome of https://github.com/doctrine/.github/pull/15

greg0ire: 4 approvals already, let's switch.

alcaeus (approved):

SenseException (approved):
@greg0ire I assume the badge will become visible after it got merged.

greg0ire:
I think you are correct, but I will check.

greg0ire:
@SenseException well… https://github.com/doctrine/collections/pull/259 :rofl:

PR #262, opened 2020-12-10
Use the ramsey/composer-install action to install dependencies

nicwortel:
Follow-up of doctrine/.github#16.

Please note: the [contributing guidelines](https://github.com/doctrine/collections/blob/1.6.x/CONTRIBUTING.md) of this repo state that you [only accepts PR's to master](https://github.com/doctrine/collections/blob/1.6.x/CONTRIBUTING.md#we-only-accept-prs--to-master), but the last 5 PR's (#261, #259, #258, #257, #256) were merged into the default branch (1.6.x). Is the `CONTRIBUTING.md` file outdated?

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -32,15 +32,8 @@ jobs:
           php-version: "${{ matrix.php-version }}"
           tools: "cs2pr"
 
-      - name: "Cache dependencies installed with Composer"
-        uses: "actions/cache@v2"
-        with:
-          path: "~/.composer/cache"
-          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
-          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
-
       - name: "Install dependencies with Composer"
-        run: "composer install --no-interaction --no-progress"
+        uses: "ramsey/composer-install@v1"
 
       # https://github.com/doctrine/.github/issues/3
       - name: "Run PHP_CodeSniffer"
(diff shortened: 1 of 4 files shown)


nicwortel:
How would you like me to handle the failing build? I see it was already failing on the default branch, but I would like to fix it in this PR so you don't have to merge a PR with a failing build.

Psalm seems to fail because of a "redundant" type cast:

```php
    /**
     * Set the number of first result that this Criteria should return.
     *
     * @param int|null $firstResult The value to set.
     *
     * @return Criteria
     */
    public function setFirstResult($firstResult)
    {
        $this->firstResult = $firstResult === null ? null : (int) $firstResult;

        return $this;
    }
```

I don't think it's possible to add a native type hint to the method parameter without breaking backward compatibility, and it seems like a good idea to force the value to `int` in case another type is provided. So I think the best solution would be to suppress `RedundantCastGivenDocblockType` errors, do you agree?

greg0ire:
> Is the CONTRIBUTING.md file outdated?

Uuuuh yes I think so, we have been migrating to merging up on all repositories, no exceptions. I think we should replace the contents of that file with a link to https://www.doctrine-project.org/contribute/index.html

greg0ire:
> it seems like a good idea to force the value to int in case another type is provided.

Although the phpdoc is not as strong a contract as a type declaration, it's still a contract IMO, and I would therefore just drop that cast. Unless that phpdoc has recently been added, a change in behavior for the case where you provide something else would not be a BC-break, because that behavior is undefined.

nicwortel:
> > it seems like a good idea to force the value to int in case another type is provided.
> 
> Although the phpdoc is not as strong a contract as a type declaration, it's still a contract IMO, and I would therefore just drop that cast. Unless that phpdoc has recently been added, a change in behavior for the case where you provide something else would not be a BC-break, because that behavior is undefined.

OK, I have removed the type cast :+1:

greg0ire (approved):

greg0ire:
Thanks @nicwortel !

PR #263, opened 2021-01-08
Allow accessing fields with a method of the same name

tszymanek:
Running a matching over a criteria lead me to an error that I cannot access the field. I then discovered the run lacks checking the object for methods with the same name as the field has. I find it a common practice to ommit the get keyword in getter methods. Thus this PR

### DIFF
--- a/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
+++ b/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
@@ -28,8 +28,8 @@ class ClosureExpressionVisitor extends ExpressionVisitor
 {
     /**
      * Accesses the field of a given object. This field has to be public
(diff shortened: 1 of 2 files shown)


malarzm:
Just for the record: this has been previously proposed in #135 and #149

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -41,6 +41,10 @@ public static function getObjectFieldValue($object, string $field)
             return $object[$field];
         }
 
+        if (method_exists($object, $field)) {
```
SenseException, line 44: `method_exists` also returns `true` if the method is not public.
https://3v4l.org/LAgYk

tszymanek: Exactly the same as it works without my changes

PR #264, opened 2021-01-14
Change Collection#slice return type to Collection

### LINKED ISSUE #195: slice() method should return a Collection instead of an array

LIQRGV:
This commit should resolve #195

### DIFF
--- a/UPGRADING-2.0.md
+++ b/UPGRADING-2.0.md
@@ -46,10 +46,11 @@ You can find a list of major changes to public API below.
 
 #### Doctrine\Common\Collections\ArrayCollection
 
-|            before           |               after               |
-|----------------------------:|:----------------------------------|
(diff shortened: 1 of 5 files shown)


LIQRGV:
Duplicate of #196

stof:
This is a BC break, so it cannot be done

SenseException:
@stof That's why 2.0 is the target branch, but I assume that there are no plans to make a collection return a new collection instance?

stof:
@SenseException this still looks like a BC break for which it is impossible to provide a smooth migration path, which makes it a very hard BC break for the ecosystem.

greg0ire:
Closing as per https://github.com/doctrine/collections/pull/264#issuecomment-1072345923

PR #269, opened 2021-02-02
Use explicit `@return` when implementing base PHP types

nicolas-grekas:
This PR replaces `{@inheritdoc}` that relate to internal PHP types by explicit `@return` annotations.

This removes the need to rely on external knowledge to get some understanding of the annotations.

Note that similar explicit annotations are already used in some places in the codebase.

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -3,6 +3,7 @@
 namespace Doctrine\Common\Collections;
 
 use Closure;
+use Traversable;
 
 /**
  * Lazy collection that is backed by a concrete collection
@@ -27,6 +28,8 @@ abstract class AbstractLazyCollection implements Collection
 
     /**
      * {@inheritDoc}
+     *
+     * @return int
      */
     public function count()
     {
@@ -275,6 +278,10 @@ abstract class AbstractLazyCollection implements Collection
 
     /**
      * {@inheritDoc}
+     *
+     * @return Traversable<int|string, mixed>
+     *
+     * @psalm-return Traversable<TKey,T>
      */
     public function getIterator()
     {
@@ -285,6 +292,8 @@ abstract class AbstractLazyCollection implements Collection
 
     /**
      * {@inheritDoc}
+     *
+     * @return bool
      *
      * @psalm-param TKey $offset
      */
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
~Please advise how I can fix the CS reports. I added `@psalm-return` annotations, but it looks like this is not enough.~

lib/Doctrine/Common/Collections/AbstractLazyCollection.php
```diff
@@ -274,7 +275,9 @@ public function slice($offset, $length = null)
     }
 
     /**
-     * {@inheritDoc}
+     * @return Traversable
```
VincentLanglet: I think `@return Traversable<int|string, mixed>` can work

nicolas-grekas: it does, thank you!

greg0ire:
Let's see if I got this right: in https://github.com/symfony/symfony/pull/30323, a new feature was introduced to Symfony, that allows to trigger a deprecation when the user extends a class without enforcing the return type documented with `@return`, because it is most likely that the package defining the class that is extended will add that return type in a next major.

One of the issues with it is that "vendor code extending other vendor code could trigger the newly introduced deprecations"

To alleviate that https://github.com/symfony/symfony/pull/33283 introduced the ability to patch vendor code to add said missing return types.

And now, what I guess this PR is about: it cannot be done for some signatures in `doctrine/collections`, because that would require using stubs like `jetbrains/phpstorm-stubs`, which is out of scope of the DebugClassLoader.

Merging this PR will effectively allow Symfony to patch vendors that extend classes from `doctrine/collections`, silencing some deprecations.

VincentLanglet:
In AbstractLazyCollection, the offsetGet method has the following phpdoc
https://github.com/doctrine/collections/blob/1b04ff044e720412994f63e51e3f206942318d07/lib/Doctrine/Common/Collections/AbstractLazyCollection.php#L298-L306

In ArrayCollection it's just https://github.com/doctrine/collections/blob/1b04ff044e720412994f63e51e3f206942318d07/lib/Doctrine/Common/Collections/ArrayCollection.php#L179-L185

The change proposed for php methods doesnt use inheritdoc at all https://github.com/doctrine/collections/pull/269/files#diff-b286fd4a4293912c0fdec02f0baaec71ee26eb6e44f54a1b025c902635096e12R168-R174

Wouldn't be better to have consistency for all the php-method of base types, either inheritdoc + phpdoc, either just phpdoc. But not having a mix of the two solutions ?

greg0ire:
I think you're right, and also I need to retarget this PR so I'll do both things.

nicolas-grekas:
You've got it right @greg0ire!

VincentLanglet (approved):

greg0ire (approved):
Looks fine especially considering the existing similar cases we have in the codebase already.

PR #274, opened 2021-06-09
PHPStan already supports array-key as template type variable bound

ondrejmirtes:
It's now even counter-productive to leave `@phpstan-template` there.

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -8,7 +8,6 @@ use Traversable;
 /**
  * Lazy collection that is backed by a concrete collection
  *
- * @phpstan-template TKey
  * @psalm-template TKey of array-key
  * @psalm-template T
  * @template-implements Collection<TKey,T>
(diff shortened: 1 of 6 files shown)


greg0ire:
We are ignoring a message right now, and it just changed:

```diff
- Array (array<TKey, T>) does not accept key int.
+ Array (array<TKey of (int|string), T>) does not accept key int.
```

That's about https://github.com/doctrine/collections/blob/94918256daa6ac99c7e5774720c0e76f01936bda/lib/Doctrine/Common/Collections/ArrayCollection.php#L312

which is described at https://github.com/doctrine/collections/blob/94918256daa6ac99c7e5774720c0e76f01936bda/lib/Doctrine/Common/Collections/ArrayCollection.php#L50-L53

I don't understand why it still fails… `array<TKey of (int|string), T>` should accept key `int`, shouldn't  it?

ondrejmirtes:
I'll update the ignore, sorry.

The error message tells you that if the user has `ArrayCollection<string, Foo>`, the method `add()` will break the key type as it inserts an `int`.

greg0ire (dismissed):
LGTM :slightly_smiling_face:

greg0ire:
:thinking: didn't notice the Psalm issue, sorry! It's pre-existing it seems.

greg0ire:
I'll ignore it in an extra commit.

greg0ire (approved):

greg0ire:
Thanks @ondrejmirtes !

ondrejmirtes:
Thank you!

PR #275, opened 2021-07-13
Update Criteria.php

Varfendell:
When using the expr function, it return `self::$expressionBuilder` which is typed |null. IDE shows possible null exeption occur. However `public static function expr()` can never return null as if self::$expressionBuilder is null, it initialize it.

The 2 solutions to manage this are:
- Type `self::$expressionBuilder` not nullable
- Type the function return type as `ExpressionBuilder`

### DIFF
--- a/lib/Doctrine/Common/Collections/Criteria.php
+++ b/lib/Doctrine/Common/Collections/Criteria.php
@@ -19,7 +19,7 @@ class Criteria
 
     public const DESC = 'DESC';
 
-    /** @var ExpressionBuilder|null */
+    /** @var ExpressionBuilder */
     private static $expressionBuilder;
 
     /** @var Expression|null */


malarzm:
@Varfendell thanks for the PR! Where (and which) is IDE showing an error? Is it when inspecting our code or when consuming `expr` method? From consumer's perspective `expr` method has a `@return` that should be honoured by IDEs, and for our internals code is perfectly fine as we do initialize it if it's null.

Varfendell:
@malarzm Phpstorm is showing this error when consuming `expr` method

Varfendell:
https://www.loom.com/share/f2646113eaa246cc833d89d5642b4721

malarzm:
This sounds like a bug in PHPStorm to me, it should respect `@return` annotation which says the `null` will not be returned

PR #277, opened 2021-07-26
Add ReturnTypeWillChange where necessary

derrabus:
PHP 8.1 will issue a deprecation for implementations of several core interfaces without a return type declaration. This PR proposes to suppress that deprecation for now because we cannot add those return types without breaking BC and bumping the PHP requirement.

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.4"
+          - "8.0"
 
     steps:
       - name: "Checkout"
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.4"
+          - "8.0"
 
     steps:
       - name: "Checkout code"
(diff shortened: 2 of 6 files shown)


phpcs.xml.dist
```diff
@@ -47,4 +47,10 @@
     <rule ref="Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps">
         <exclude-pattern>tests/Doctrine/Tests/Common/Collections/ClosureExpressionVisitorTest.php</exclude-pattern>
     </rule>
+
+    <rule ref="SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse">
+        <!-- This rule causes false-positives for imported attribute classes. -->
+        <exclude-pattern>lib/Doctrine/Common/Collections/ArrayCollection.php</exclude-pattern>
+        <exclude-pattern>lib/Doctrine/Common/Collections/AbstractLazyCollection.php</exclude-pattern>
+    </rule>
```
derrabus: This can be reverted if we merge #278.

derrabus:
@greg0ire Sorry for the direct ping, but merging this PR would remove some noise from our CI. Can you have a look, please?

phpcs.xml.dist
greg0ire: Done

greg0ire (approved):

greg0ire:
Thanks @derrabus !

SenseException (approved):

PR #280, opened 2021-08-10
Add explicit `@return` type next to `#[ReturnTypeWillChange]`

nicolas-grekas:
Similar to #269 but taking types added to PHP 8.1 into account.


alcaeus (approved):

alcaeus:
Thanks @nicolas-grekas!

PR #289, opened 2021-10-20
Admit no prefix as getter

jorge-ruukfy:
As an alternative to reflection or closure hacking to properly get the private property whenever not using `getXXX()` `isXXX()` accessors,
just added blank prefix '' to call DDD oriented notation (ex. `id()` )

Is adverted to who want to use this in a future, that if the method returns a valueObject, to use ::CONTAINS operator that casts to string, although the VO must be stringable

### DIFF
--- a/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
+++ b/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
@@ -42,7 +42,7 @@ class ClosureExpressionVisitor extends ExpressionVisitor
             return $object[$field];
         }
 
-        $accessors = ['get', 'is'];
+        $accessors = ['get', 'is', ''];
 
         foreach ($accessors as $accessor) {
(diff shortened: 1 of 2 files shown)


SenseException:
Interesting addition prevent classic getter names. Can you please also extend the tests for this naming?

SenseException:
Please fix the issues listed in the coding standards results of the build.

SenseException:
Looks great. Please adapt your 2 last commit message as described in our contribution guideline. https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches

jorge-ruukfy:
@SenseException :tada:

SenseException (dismissed):

derrabus:
Do we consider this to be a bugfix? Otherwise, we should target 1.7.x, shouldn't we?

SenseException:
You're right. This is a feature.

derrabus (approved):

PR #290, opened 2021-11-01
Bump PHPStan level

derrabus:
This PR bumps the PHPStan level to 8 and fixes some issues found on the way.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     "require-dev": {
         "phpunit/phpunit": "^7.5 || ^8.5 || ^9.1.5",
         "doctrine/coding-standard": "^9.0",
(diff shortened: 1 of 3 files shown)


greg0ire (commented):
I noticed that you switched to tabs on the `doctrine/persistence` PR but not on this one, is it deliberate?

greg0ire (approved):
Ah actually it was just for the baseline. This should be fine.

greg0ire:
Thanks @derrabus !

derrabus:
> I noticed that you switched to tabs on the `doctrine/persistence` PR but not on this one, is it deliberate?

I have let PHPStan regenerate the baseline. I don't think we should edit that file manually.

PR #296, opened 2022-03-06
Psalm 4.22 and PHPStan 1.4.8

franmomu:


lib/Doctrine/Common/Collections/AbstractLazyCollection.php
```diff
@@ -318,9 +318,12 @@ public function offsetGet($offset)
     }
 
     /**
-     * {@inheritDoc}
+     * @param TKey|null $offset
+     * @param T         $value
      *
      * @return void
+     *
+     * @psalm-suppress PossiblyNullArgument
```
franmomu: This is because:

```
ERROR: PossiblyNullArgument - lib/Doctrine/Common/Collections/AbstractLazyCollection.php:330:38 - Argument 1 of Doctrine\Common\Collections\Collection::offsetSet cannot be null, possibly null value provided (see https://psalm.dev/078)
        $this->collection->offsetSet($offset, $value);
```

Maybe `ArrayAccess` should explicitly say that `$offset` could be `TKey|null`, I'll try to take a look

derrabus: Please ignore this error via the psalm.xml.dist file and add a link to your Psalm issue as a comment.

psalm.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0"?>
+<psalm
```
derrabus: Why did you rename the file?

franmomu: I understand `.dist` files as generic files without specific configuration shipped with a library and once they have custom configuration you can remove that suffix. In this case this configuration file is specific for this package and it's not supposed to be used by someone else.

Anyway, I can append the `.dist` suffix if that is not right (or do it in another PR).

derrabus: Yes, please add it back.

derrabus (approved):

PR #301, opened 2022-07-13
Add toList function

jack-worman:
Adds `Collection::toList(): list<T>`

In my own projects, I have been having to do the following pattern to make static analysis happy:
```php
/** @var list<T> $list */
$list = $listCollection->toArray();
```

It would be nice not to have to use doc-block annotations.

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -158,6 +158,16 @@ abstract class AbstractLazyCollection implements Collection
     }
 
     /**
+     * @return list<T>
+     */
+    public function toList(): array
+    {
+        $this->initialize();
+
+        return $this->collection->toList();
+    }
+
+    /**
      * {@inheritDoc}
      */
     public function first()
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -72,6 +72,17 @@ class ArrayCollection implements Collection, Selectable
     }
 
     /**
+     * @return list<T>
+     */
+    public function toList(): array
+    {
+        if (\PHP_VERSION_ID >= 80100 && \array_is_list($this->elements)) {
+            return $this->elements;
+        }
+        return \array_values($this->elements);
+    }
+
+    /**
      * {@inheritDoc}
      */
     public function first()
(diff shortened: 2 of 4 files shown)


greg0ire:
Retargeting to 1.7.x since this is a new feature. Please rebase on that branch and get rid of the extra commits.

jack-worman:
> 

> Retargeting to 1.7.x since this is a new feature. Please rebase on that branch and get rid of the extra commits

Done

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -28,6 +28,8 @@
  * @psalm-template T
  * @template-extends IteratorAggregate<TKey, T>
  * @template-extends ArrayAccess<TKey, T>
+ *
+ * @method array toList()
```
derrabus, line 32:
```suggestion
 * @method list<T> toList()
```

lib/Doctrine/Common/Collections/AbstractLazyCollection.php
```diff
@@ -157,6 +157,16 @@ public function toArray()
...
     }
 
+    /**
+     * @return list<T>
+     */
+    public function toList(): array
+    {
+        $this->initialize();
+
+        return $this->collection->toList();
```
derrabus, line 167: We cannot be certain that the wrapped collection already implements the new method.

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -71,6 +71,17 @@ public function toArray()
...
     }
 
+    /**
+     * @return list<T>
+     */
+    public function toList(): array
+    {
+        if (\PHP_VERSION_ID >= 80100 && \array_is_list($this->elements)) {
+            return $this->elements;
+        }
```
derrabus, line 81: Did you benchmark this code? Is `array_is_list()` really so much faster than `array_values()`? If not, I'd remove this block.

derrabus (changes requested):

malarzm (changes requested):
What's the point of new method when https://github.com/doctrine/collections/blob/c1ec9823f9250fefb274cccf62970b5c167e6bc7/lib/Doctrine/Common/Collections/ArrayCollection.php#L281-L284 is already in place?

jack-worman:
> What's the point of new method when
> 
> https://github.com/doctrine/collections/blob/c1ec9823f9250fefb274cccf62970b5c167e6bc7/lib/Doctrine/Common/Collections/ArrayCollection.php#L281-L284

That is a big face-palm.... I guess I will close this PR. Thanks!

PR #305, opened 2022-08-02
Drop old deps

greg0ire:
Note: I haven't used Rector (yet). Type additions were made thanks to phpcbf.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,14 +33,14 @@
     ],
     "homepage": "https://www.doctrine-project.org/projects/collections.html",
     "require": {
-        "php": "^7.2 || ^8.0"
+        "php": "^8.1"
     },
     "require-dev": {
         "ext-json": "*",
         "doctrine/coding-standard": "^9.0",
-        "phpstan/phpstan": "^1.4.8",
+        "phpstan/phpstan": "^1.8",
         "phpstan/phpstan-phpunit": "^1.0",
-        "phpunit/phpunit": "^7.5 || ^8.5 || ^9.1.5",
+        "phpunit/phpunit": "^9.5",
         "vimeo/psalm": "^4.22"
     },
     "autoload": {
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -275,7 +275,7 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * @return int|string|bool The key/index of the element or FALSE if the element was not found.
      * @psalm-return TKey|false
      */
-    public function indexOf(mixed $element): int|string|bool;
+    public function indexOf(mixed $element): int|string|false;
 
     /**
      * Extracts a slice of $length elements starting at position $offset from the Collection.
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -41,7 +41,7 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      *
      * @return true Always TRUE.
      */
-    public function add($element): bool;
+    public function add(mixed $element): bool;
 
     /**
      * Clears the collection, removing all elements.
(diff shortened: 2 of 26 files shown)


composer.json
```diff
@@ -33,14 +33,14 @@
...
-        "php": "^7.2 || ^8.0"
+        "php": "^8.1"
     },
     "require-dev": {
         "ext-json": "*",
         "doctrine/coding-standard": "^9.0",
         "phpstan/phpstan": "^1.4.8",
         "phpstan/phpstan-phpunit": "^1.0",
-        "phpunit/phpunit": "^7.5 || ^8.5 || ^9.1.5",
+        "phpunit/phpunit": "^9.1.5",
```
derrabus:
```suggestion
        "phpunit/phpunit": "^9.5",
```

PHPUnit 9.4 and below won't work on PHP 8.1.

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -283,7 +275,7 @@ public function partition(Closure $p): array;
      * @return int|string|bool The key/index of the element or FALSE if the element was not found.
      * @psalm-return TKey|false
      */
-    public function indexOf($element);
+    public function indexOf(mixed $element): int|string|bool;
```
derrabus:
```suggestion
    public function indexOf(mixed $element): int|string|false;
```

We never return `true`, do we?

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -41,7 +41,7 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      *
      * @return true Always TRUE.
      */
-    public function add($element): bool;
+    public function add(mixed $element): bool;
```
derrabus, line 44: Shall we change the return type to `void`?

greg0ire: Yes 👍 Always returning `true` looks weird.

greg0ire:
I think it's best to leave Rector to a subsequent PR so that it stays relatively small.

phpcs.xml.dist
```diff
@@ -6,24 +6,15 @@
...
 
     <file>lib</file>
     <file>tests</file>
 
-    <rule ref="Doctrine">
-        <exclude name="SlevomatCodingStandard.TypeHints.DeclareStrictTypes"/>
-        <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint"/>
-        <exclude name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint"/>
-
-        <!-- Since Collections are built around mixed content, this has no value here -->
```
greg0ire, line 22: That statement might be outdated: removing this block did not force me to add extra type declarations, probably thanks to the existing generic annotations.

lib/Doctrine/Common/Collections/Collection.php
malarzm: I don't think it's a good idea to change return types between 1.x and 2.x. How are we going to support both versions then?

greg0ire: Good point! A solution could be to have different traits with different signatures and conditionally use one trait or the other.

derrabus (dismissed):

lib/Doctrine/Common/Collections/Collection.php
malarzm: That will still break the code at large. From our own org we have ODM's dynamic collection generation (which uses trait for generated PersistentCollection so it's ok-ish) but effectively it allowed users to have their own collection classes in documents. I'm myself maintaing https://github.com/malarzm/collections which provides own implementations of `Collection` interface.

My point is: if we require everybody with own implementations of the interface to have traits, adoption will be slow. Always returning `true` may be weird, but bool -> void is too much of a change for a package so widely used.

derrabus: Good point. `Collection` is an interface that should be implemented in userland and the change I suggested would cause too much trouble in existing implementations for too little gain. Let's revert.

greg0ire: Reverted. We can deal with this separately, for instance we could drop the return type declaration in 2.0.x, then add `void` in 3.0.x.

derrabus (approved):

PR #209, opened 2019-10-17
Initialize $element property of ArrayCollection

FabienSalles:
Initialize the property $element to an Array in order to avoid warning when we unserialize a serialized array (see #155) :

`Warning: array_values() expects parameter 1 to be array, null given`

PR #306, opened 2022-08-03
Run Rector

greg0ire:
This is what happens when you run rector with the following configuration:

```php
<?php

declare(strict_types=1);

use Rector\Config\RectorConfig;
use Rector\Php71\Rector\FuncCall\CountOnNullRector;
use Rector\Set\ValueObject\LevelSetList;
use Rector\TypeDeclaration\Rector\FunctionLike\ParamTypeDeclarationRector;
use Rector\TypeDeclaration\Rector\FunctionLike\ReturnTypeDeclarationRector;

return static function (RectorConfig $rectorConfig): void {
    $rectorConfig->paths([
        __DIR__ . '/lib',
        __DIR__ . '/tests',
    ]);

    $rectorConfig->sets([
        LevelSetList::UP_TO_PHP_81,
    ]);

    $rectorConfig->rule(ParamTypeDeclarationRector::class);
    $rectorConfig->rule(ReturnTypeDeclarationRector::class);

    $rectorConfig->importNames();
    $rectorConfig->skip([
        CountOnNullRector::class, // I haven't reevaluated this rule
    ]);
};
```

I made some manual tweaks, for instance for CPP, I moved / will move phpdoc comments to `@param` rather than `@var` on properties.

### DIFF
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -7,6 +7,7 @@ namespace Doctrine\Common\Collections;
 use ArrayIterator;
 use Closure;
 use Doctrine\Common\Collections\Expr\ClosureExpressionVisitor;
+use Stringable;
 use Traversable;
 
 use function array_filter;
@@ -44,7 +45,7 @@ use const ARRAY_FILTER_USE_BOTH;
  * @template-implements Selectable<TKey,T>
  * @psalm-consistent-constructor
  */
-class ArrayCollection implements Collection, Selectable
+class ArrayCollection implements Collection, Selectable, Stringable
 {
     /**
      * An array containing the entries of this collection.
@@ -87,7 +88,6 @@ class ArrayCollection implements Collection, Selectable
      * @param array $elements Elements.
      * @psalm-param array<K,V> $elements
      *
-     * @return static
      * @psalm-return static<K,V>
      *
      * @psalm-template K of array-key
--- a/lib/Doctrine/Common/Collections/Criteria.php
+++ b/lib/Doctrine/Common/Collections/Criteria.php
@@ -17,26 +17,21 @@ use function strtoupper;
  */
 class Criteria
 {
-    public const ASC  = 'ASC';
-    public const DESC = 'DESC';
+    final public const ASC  = 'ASC';
+    final public const DESC = 'DESC';
 
     private static ?ExpressionBuilder $expressionBuilder = null;
-
-    private ?Expression $expression = null;
 
     /** @var array<string, string> */
     private array $orderings = [];
 
     private ?int $firstResult = null;
-
-    private ?int $maxResults = null;
+    private ?int $maxResults  = null;
 
     /**
      * Creates an instance of the class.
-     *
-     * @return Criteria
      */
-    public static function create(): self
+    public static function create(): static
     {
         return new static();
     }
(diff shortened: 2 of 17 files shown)


lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -44,25 +45,17 @@
...
      * Initializes a new ArrayCollection.
      *
      * @param array $elements
      * @psalm-param array<TKey,T> $elements
      */
-    public function __construct(array $elements = [])
-    {
-        $this->elements = $elements;
+    public function __construct(
+        private array $elements = []
```
greg0ire: This seems to cause an issue when unserializing an empty array collection. The failing test was introduced in #209

greg0ire: Let's revert I guess.

lib/Doctrine/Common/Collections/Criteria.php
```diff
@@ -17,26 +17,21 @@
...
-    private ?int $maxResults = null;
+    private ?int $maxResults  = null;
 
     /**
      * Creates an instance of the class.
-     *
-     * @return Criteria
      */
-    public static function create(): self
+    public static function create(): static
```
malarzm, line 34: Shall we change self -> static across the file?

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -110,116 +106,68 @@ public static function sortByField(string $name, int $orientation = 1, ?Closure
         };
     }
 
-    public function walkComparison(Comparison $comparison): mixed
+    public function walkComparison(Comparison $comparison): Closure
```
malarzm, line 117: I think we should start documenting BC breaks along with PRs, there'll be no easy way by the end of the journey

malarzm (dismissed):

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
greg0ire: There is a nice upgrade guide already, I'll update it.

lib/Doctrine/Common/Collections/Criteria.php
```diff
@@ -74,10 +71,8 @@ public function __construct(?Expression $expression = null, ?array $orderings =
 
     /**
      * Sets the where expression to evaluate when this Criteria is searched for.
-     *
-     * @return Criteria
```
derrabus: This should be `@return $this` instead.

greg0ire: You're right, but that should be contributed to 1.7.x

greg0ire: See #307

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -44,7 +45,7 @@
  * @template-implements Selectable<TKey,T>
  * @psalm-consistent-constructor
  */
-class ArrayCollection implements Collection, Selectable
+class ArrayCollection implements Collection, Selectable, Stringable
```
derrabus, line 48: I wasn't aware that `ArrayCollection`s could be cast to string and I must admit, I find the implementation rather surprising. 😅

lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php
```diff
@@ -118,116 +114,68 @@ public static function sortByField(string $name, int $orientation = 1, ?Closure
...
+        };
     }
 
     public function walkValue(Value $value): mixed
     {
         return $value->getValue();
     }
 
-    public function walkCompositeExpression(CompositeExpression $expr): mixed
+    public function walkCompositeExpression(CompositeExpression $expr): callable
```
derrabus:
```suggestion
    public function walkCompositeExpression(CompositeExpression $expr): Closure
```

We could narrow the type here as well.

derrabus (dismissed):

derrabus (approved):

PR #311, opened 2022-08-03
Deprecate passing null as offset

derrabus:
In order to be able to narrow down parameter and return types for 2.0, I think we should disallow using `null` as offset for `Criteria` instances. We've applied similar changes to DBAL and ORM already.

### DIFF
--- a/lib/Doctrine/Common/Collections/Criteria.php
+++ b/lib/Doctrine/Common/Collections/Criteria.php
@@ -4,8 +4,10 @@ namespace Doctrine\Common\Collections;
 
 use Doctrine\Common\Collections\Expr\CompositeExpression;
 use Doctrine\Common\Collections\Expr\Expression;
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Common/Collections/Criteria.php
```diff
@@ -69,6 +71,15 @@ public function __construct(?Expression $expression = null, ?array $orderings =
     {
         $this->expression = $expression;
 
+        if ($firstResult === null && func_num_args() > 2) {
+            Deprecation::trigger(
+                'doctrine/collections',
+                'TODO',
```
greg0ire: :point_up: :trollface:

derrabus: 😓😓😓

greg0ire (approved):

PR #313, opened 2022-08-08
Update phpdoc of the AbstractLazyCollection

VincentLanglet:
When extending the AbstractLazyCollection, I'm getting an error PropertyNotSetInConstructor from psalm.
This is because the `collection` property is declared as a Collection but not set in the constructor of my class.

I personally initialize and set the collection in the `doInitialize` method, and I thought this was one of the purpose of this class. If I understand correctly it meant that it's allowed to have a `collection` property nullable then.

I made the change in the phpdoc related to this.

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -7,8 +7,6 @@ use LogicException;
 use ReturnTypeWillChange;
 use Traversable;
 
-use function assert;
-
 /**
  * Lazy collection that is backed by a concrete collection
  *
@@ -38,7 +36,6 @@ abstract class AbstractLazyCollection implements Collection
     public function count()
     {
         $this->initialize();
-        assert($this->collection !== null);
 
         return $this->collection->count();
     }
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Collections/AbstractLazyCollection.php
```diff
@@ -35,6 +38,7 @@ abstract class AbstractLazyCollection implements Collection
     public function count()
     {
         $this->initialize();
+        assert($this->collection !== null);
```
derrabus: Do we really need to C&P this everywhere? Calling the `initialize()` method already makes sure that the collection is initialized.

VincentLanglet: It's for static analysis.

Adding 
```
@psalm-assert Collection<TKey,T> $this->collection
```
works for psalm, but phpstan doesn't understand this tag yet.

So it reports
```
 Cannot call method offsetUnset() on Doctrine\Common\Collections\Collection<TKey of (int|string), T>|null.
```
But I can add the pattern to the ignoreError section, until it's supported by phpstan (which should happen soon).

VincentLanglet: I updated the PR

derrabus: That works for me, thanks.

derrabus (approved):

malarzm:
Thanks @VincentLanglet!

Issue #291, opened 2021-11-10
Question: `Selectable` return value


boesing:
Hey there,

as of now, I do only know about `Selectable#matching` returns `Collection` (which also includes `Selectable`).
So when we use this in our projects, we almost always do also assert that the returned value implements both `Collection&Selectable` (which is actually true for `ArrayCollection` in ORM context tho).

So I wondered if there are plans to either annotate something like:

```
interface Selectable
{
    /**
     * Selects all elements from a selectable that match the expression and
     * returns a new collection containing these elements.
     *
     * @return Collection<mixed>
     * @psalm-return Collection<TKey,T>&Selectable<TKey,T>
     */
    public function matching(Criteria $criteria);
}
```

or if the `Selectable` interface also extends the `Collection` interface.

----

So why do I ask for this?

We do have some use-cases where we do pre-filter collections form within a method.
So code like this is something which is "normal" in our codebase:

```
final class Foo
{
    /** @param Collection<array-key,Entity>&Selectable<array-key,Entity> */
    public function __construct(Collection $foo)
    {
        $this->foo = $foo;
    }

    /** @return Collection<array-key,Entity>&Selectable<array-key,Entity> */
    public function getValidEntries(DateTimeImmutable $from): Collection
    {
        $criteria = Criteria::create();
        $expression = Criteria::expr();
        return $this->foo->matching($expression->gte('validFrom', $from));
    }
   
}
```

Afterwards we do want to apply additional criterias (or not) so we need another `Selectable` (which is actually not returned by `Selectable#matching`).

Is there any best practice here?

AndrolGenhald:
I run into this often enough with Psalm that I've stopped suppressing it (which risks suppressing an actual problem) and started using this utility function to wrap the result:

```php
/**
 * Workaround for https://github.com/doctrine/collections/issues/291
 *
 * @template TKey of array-key
 * @template TValue
 * @template TCollection of Collection<TKey, TValue>
 * @param TCollection $arg
 *
 * @return TCollection&Selectable<TKey, TValue>
 */
function assertSelectable($arg)
{
    if (!$arg instanceof Selectable) {
        throw new \InvalidArgumentException("Argument must be Selectable");
    }

    return $arg;
}
```

It would be nice if this could be fixed here somehow. I'd rather not just stub `Selectable` to change the return type since it's possible to return a `Collection` that doesn't implement `Selectable`.

greg0ire:
[...]

PR #314, opened 2022-08-17
Make `Selectable#matching` return `Collection`

boesing:
Hey there,

as stated in #291, `Selectable#matching` always returns a `Collection` which is also `Selectable`.
Ping @greg0ire @AndrolGenhald

Happy to get feedback here. 
Since we do not change any code here but adding an additional return value (which is already the case in almost every implementation), I've targeted 1.6.x rather than 1.7.x

Let me know if this should be released in a new minor instead, so I can rebase the branch appropriately.

Thanks!

### DIFF
--- a/tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
+++ b/tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
@@ -3,6 +3,8 @@
 namespace Doctrine\Tests\Common\Collections;
 
 use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\Common\Collections\Selectable;
 use PHPUnit\Framework\TestCase;
 use stdClass;
 
@@ -10,6 +12,7 @@ use function count;
 use function is_array;
 use function is_numeric;
 use function is_string;
+use function sprintf;
 
 abstract class BaseCollectionTest extends TestCase
 {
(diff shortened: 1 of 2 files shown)


greg0ire:
Please rebase on 1.7.x, you're not making the phpdoc more accurate, you are making it more precise.

boesing:
I rebased against 1.7.x

FYI: I realized that there are some commits in 1.6.x which were not merged-up to 1.7.x

derrabus:
Thanks, I'll fix that in #315.

derrabus:
Can you have a look at the PHPCS failure?

tests/Doctrine/Tests/Common/Collections/BaseCollectionTest.php
```diff
@@ -249,5 +249,6 @@ public function testMatchingAlwaysReturnsCollection(): void
         $criteria = Criteria::create();
 
         self::assertInstanceOf(Collection::class, $this->collection->matching($criteria));
+        self::assertInstanceOf(Selectable::class, $this->collection->matching($criteria));
```
greg0ire, line 253: We don't do semantic commits here but anyway, why did you prefix this one with `bugfix:` ? It only modifies a test…

boesing: it was a bug in the test tho. I did not assert what I wanted to assert. Feel free to squash this PR tho.

boesing: Squashed my commits now.

boesing:
> Can you have a look at the PHPCS failure?

done

derrabus (approved):

greg0ire:
Thanks @boesing !

PR #317, opened 2022-08-18
Update branch metadata

greg0ire:
I noticed that branches prior to 1.6.x don't exist. Should we delete 1.6.x?

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -11,16 +11,22 @@
             "upcoming": true
         },
         {
+            "name": "1.8",
+            "branchName": "1.8.x",
+            "slug": "1.8",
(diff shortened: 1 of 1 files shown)


SenseException:
Why removing 1.6? Older versions are usually kept in the docs.

greg0ire:
I didn't realize there never were docs for 1.5.x: #256 Let me restore the doc config for 1.6.x

SenseException (approved):
For removing a version we should also keep in mind that some Google top search results to the docs are on older versions. It would take some time before the 404 will be removed from the results and I don't know if the next versions would succeed. The Algolia indices to the removed versions also still exist.

PR #318, opened 2022-08-22
Add psalm-assert-if-true to `AbstractLazyCollection::isInitialized`

VincentLanglet:
Related to https://github.com/doctrine/collections/pull/313 @derrabus 

It should solve some of PossiblyNull issues of doctrine/orm https://github.com/doctrine/orm/runs/7947119203?check_suite_focus=true

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -347,6 +347,8 @@ abstract class AbstractLazyCollection implements Collection
      * Is the lazy collection already initialized?
      *
      * @return bool
+     *
+     * @psalm-assert-if-true Collection<TKey,T> $this->collection
      */
     public function isInitialized()
     {


greg0ire (approved):

SenseException:
It looks correct, but I can't make psalm fail for this assertion in case of a type mismatch on true.

SenseException:
It looks correct, but I can't make psalm fail for this assertion in case of a type mismatch on true for `isInitialized()`.

greg0ire:
@SenseException I don't understand, what do you mean by that? Can you please elaborate?

SenseException:
@greg0ire When I add something, like e.g. a new working test, I try to make it fail to see if it's really covering the case of mine. Similar to what mutation tests do. The same goes for additions in code analysis: I try to make it fail to see if a psalm annotation is actually doing something.

I was unable to make psalm fail to see if the addition of this PR is actually doing its job.

greg0ire:
OK, if you change the body of `AbstractLazyCollection::count()` to the following, you will see that this PR is useful:

```php
        if ($this->isInitialized()) {
            return $this->collection->count();
        }

        return 0;

```

SenseException (approved):
Thanks. I was able to reproduce the benefit of this PR.

greg0ire:
Thanks @VincentLanglet !

PR #320, opened 2022-08-22
Remove remnants from Infection

greg0ire:
Looks like it got lost when migrating to Github actions and removed from
composer.json during a merge up.

I think it would be hard to reintroduce it, and that that should not be
done on the next major branch in any case.


SenseException (approved):
I didn't know that infection was used in a Doctrine project.

greg0ire:
Neither did I! I think it might be a welcome addition, but that's substantial work.

Issue #117, opened 2017-05-02
question regarding Closure


bashilbers:
Hi,

Why does the `Collection` interface depend `Closure` (e.g. for the `fillter` method) instead of `callable`?

Just a simple usecase: using classes with `__invoke` to filter a collection.

(happy to help here)

lcobucci:
@bashilbers I'm not sure but it sounds like for historical reasons (and BC).

`Closure` was added in PHP 5.3 and `callable` in PHP 5.4 and changing the type-hint on the interface would be a BC-break...

We can think of doing that for our next major release, though.

bashilbers:
Makes sense. Is there a roadmap? I would like to contribute for the next major release.

Ocramius:
No roadmap for collections yet: as far as I can see, changing collections
is a huge issue, so it will likely stay as-is while we develop a new
package with a different API instead...

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

On Wed, May 3, 2017 at 8:48 AM, Sebastiaan Hilbers <notifications@github.com

bashilbers:
Closing this due to the reason that BC forces it to stay the way it is now

PR #321, opened 2022-08-24
Change Closure to callable

VincentLanglet:
Close https://github.com/doctrine/collections/issues/231

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -20,6 +20,21 @@ types. This would only work in PHP 7.2+ which is the first version featuring
 
 You can find a list of major changes to public API below.
 
+### Closure vs callable
+
+While 1.x branch is using `Closure` as param type declaration, 2.x is using `callable`.
+If you want to support both branches you can use `Closure|callable` (or no type declaration at all).
+
+```php
+class MyCollection implements Collection
+{
+    public function exists(Closure|callable $p): bool
+    {
+        // ...
+    }
+}
+```
+
 ### Doctrine\Common\Collections\Collection
 
 |             before             |                  after                         |
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Doctrine\Common\Collections;
 
-use Closure;
 use LogicException;
 use Traversable;
 
@@ -172,14 +171,14 @@ abstract class AbstractLazyCollection implements Collection
         return $this->collection->next();
     }
 
-    public function exists(Closure $p): bool
+    public function exists(callable $p): bool
     {
         $this->initialize();
 
         return $this->collection->exists($p);
     }
 
-    public function findFirst(Closure $p): mixed
+    public function findFirst(callable $p): mixed
     {
         $this->initialize();
 
@@ -187,19 +186,19 @@ abstract class AbstractLazyCollection implements Collection
     }
 
     /**
-     * @psalm-param Closure(T=, TKey=):bool $p
+     * @psalm-param callable(T=, TKey=):bool $p
      *
      * @return Collection<mixed>
      * @psalm-return Collection<TKey, T>
      */
-    public function filter(Closure $p): Collection
+    public function filter(callable $p): Collection
     {
         $this->initialize();
 
         return $this->collection->filter($p);
     }
 
-    public function forAll(Closure $p): bool
+    public function forAll(callable $p): bool
     {
         $this->initialize();
 
@@ -207,14 +206,14 @@ abstract class AbstractLazyCollection implements Collection
     }
 
     /**
-     * @psalm-param Closure(T=):U $func
+     * @psalm-param callable(T=):U $func
      *
      * @return Collection<mixed>
      * @psalm-return Collection<TKey, U>
      *
      * @psalm-template U
      */
-    public function map(Closure $func): Collection
+    public function map(callable $func): Collection
     {
         $this->initialize();
 
@@ -224,7 +223,7 @@ abstract class AbstractLazyCollection implements Collection
     /**
      * {@inheritDoc}
      */
-    public function reduce(Closure $func, $initial = null): mixed
+    public function reduce(callable $func, $initial = null): mixed
     {
         $this->initialize();
 
@@ -234,7 +233,7 @@ abstract class AbstractLazyCollection implements Collection
     /**
      * {@inheritDoc}
      */
-    public function partition(Closure $p): array
+    public function partition(callable $p): array
     {
         $this->initialize();
 
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Doctrine\Common\Collections;
 
 use ArrayIterator;
-use Closure;
 use Doctrine\Common\Collections\Expr\ClosureExpressionVisitor;
 use Stringable;
 use Traversable;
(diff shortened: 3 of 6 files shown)


VincentLanglet:
Psalm failing because of https://github.com/vimeo/psalm/pull/8435 @greg0ire
Should I add @psalm-suppress ?

greg0ire:
We prefer centralized ignore rules. Please add the link above when adding your ignore rule.

VincentLanglet:
> We prefer centralized ignore rules. Please add the link above when adding your ignore rule.

Done

greg0ire:
Sorry, I meant centralized in the config file directly:

https://github.com/doctrine/collections/blob/d93dbc5ef2e6e98a5b16ca5b07c3e15de9fd9113/psalm.xml.dist#L16-L40

If you use a baseline file, chances are the links you add are going to go away when using a command to re-generate the baseline.

VincentLanglet:
Fixed

but IMHO psalm-suppress are the best because as soon as the version is released the psalm suppress is reported as useless, so you dont forget.

greg0ire:
> the psalm suppress is reported as useless, so you dont forget

That's great, I didn't know about this! I wonder why it does not happen when using a config file.

psalm-baseline.xml
```diff
@@ -0,0 +1,12 @@
...
+  <!-- https://github.com/vimeo/psalm/pull/8435 -->
+  <file src="lib/Doctrine/Common/Collections/ArrayCollection.php">
+    <MixedInferredReturnType occurrences="1">
+      <code>mixed</code>
+    </MixedInferredReturnType>
+    <MixedReturnStatement occurrences="1">
+      <code>array_reduce($this-&gt;elements, $func, $initial)</code>
+    </MixedReturnStatement>
+  </file>
+</files>
```
greg0ire: I think you forgot to remove this file.

greg0ire:
Can you please add an entry to the upgrade file describing how to write an implementation of `Collection` that is compatible with v1 and v2? Right now this is the solution I see: https://3v4l.org/5aLXQ Using just `callable` doesn't work, it does not seem that `Closure` "extends" `callable`: https://3v4l.org/EZVGv

UPGRADE.md
```diff
@@ -20,6 +20,21 @@ types. This would only work in PHP 7.2+ which is the first version featuring
 
 You can find a list of major changes to public API below.
 
+### Closure vs callable
+
+While 1.x branch is using `Closure` as param typehint, 2.x is using `callable`.
+If you want to support both branch you'll have to use `Closure|callable`.
```
greg0ire:
```suggestion
If you want to support both branches you can use `Closure|callable` (or no type declaration at all).
```

UPGRADE.md
```diff
@@ -20,6 +20,21 @@ types. This would only work in PHP 7.2+ which is the first version featuring
 
 You can find a list of major changes to public API below.
 
+### Closure vs callable
+
+While 1.x branch is using `Closure` as param typehint, 2.x is using `callable`.
```
greg0ire:
```suggestion
While 1.x branch is using `Closure` as param type declaration, 2.x is using `callable`.
```

greg0ire (commented):
A few nitpicks

greg0ire (approved):

derrabus (changes requested):
Is the issue we're fixing here still relevant in PHP 8.1 and first-class callable syntax? I'd rather drop support for `callable`s other than `Closure`s everywhere tbh.

VincentLanglet:
> Is the issue we're fixing here still relevant in PHP 8.1 and first-class callable syntax? I'd rather drop support for `callable`s other than `Closure`s everywhere tbh.

I don't understand.

Every closure are callable. So you're not dropping support for Closure.
First-class callable doesn't solve the issue that these method could support all the callable.

```
function foo(Closure $call) {}
function bar(callable $call) {}

class A
{
	public function __invoke() {}
}

foo(strlen(...)); // works
bar(strlen(...));  // works

foo(fn () => 1); // works
bar(fn () => 1); // works

foo(new A()); // don't work
bar(new A()); // works
```

greg0ire:
I didn't even know about first-class callable syntax, but now I think @derrabus point is: should we encourage usage of ugly stuff like `'strlen'` when our consumers have `strlen(...)`, since 8.1 is required:

https://github.com/doctrine/collections/blob/cec2511e62d35e41d56cd97858ec64de3365eb4b/composer.json#L36 

> So you're not dropping support for Closure.

@derrabus wasn't suggesting your PR did that. I think you missed the word "other" in his answer.

VincentLanglet:
> I didn't even know about first-class callable syntax, but now I think @derrabus point is: should we encourage usage of ugly stuff like `'strlen'` when our consumers have `strlen(...)`, since 8.1 is required:
> 
> https://github.com/doctrine/collections/blob/cec2511e62d35e41d56cd97858ec64de3365eb4b/composer.json#L36
> 
> > So you're not dropping support for Closure.

I don't really see the point about restricting the param by ideology if the method can support it.
The implementation doesn't need the param to specifically to be a closure and not a callable.
If `foo('strlen')` works correctly, there is no reason to restrict people to write `foo(strlen(...))`.

It's acting like `callable` were deprecated by php but it's not.
A php rfc should be proposed to deprecate callable before refusing support for it.

VincentLanglet:
> I didn't even know about first-class callable syntax, but now I think @derrabus point is: should we encourage usage of ugly stuff like 'strlen' when our consumers have strlen(...), since 8.1 is required

Also that doesn't solve the issue for Class with `__invoke` method.
Cf https://github.com/doctrine/collections/issues/117#issue-225716241

greg0ire:
In that case, maybe it's better to merge this and let PHP itself deprecate `'strlen'` in a future major release.

VincentLanglet:
> In that case, maybe it's better to merge this and let PHP itself deprecate `'strlen'` in a future major release.

Even if it's deprecated, callable will still be needed for invokable classes, no ?

greg0ire:
I suppose so.

derrabus:
The `callable` pseudo-type is ugly, especially the string or array kind. Unlike other types, the `callable` type is scope-dependant: Depending on when and where you look at it, the same array or string can be `callable` or not. As a consequence, `callable` cannot be used as property type. Moreover, on a hot path, a closure will always perform better than an array callable or a string callable. Thus, widening all interfaces to `callable` forces implementations to either live with more edge-cases or covert to `Closure` immediately everywhere.

Since with the new first-class callable syntax, creating `Closure` objects from existing callables has become a piece of cake, I fail to see what value the proposed change would provide on PHP 8.1.

Note that we're not simply building a brand-new API here: The PR proposes to change existing and established interfaces. A downstream implementation that wants to support v1 and v2 at the same time is either forces to drop parameter types on closure parameters completely or drop support for PHP 7 and use an ugly union type. If we were to pursue that path, it should be worth it, and I believe it's not.

---

> callable will still be needed for invokable classes

```php
a_function_that_expects_a_closure($myInvokableObject(...));
a_function_that_expects_a_closure((new MyInvokableClass)(...));
```

---

I understand that we would improve the situation for invokable objects a little by widening to `callable`. But that's still too little gain for too much pain, imho.

I appreciate the effort you've put in this PR and I'm honestly sorry that I've jumped onto this train that late. But I'm genuinely not convinced that this change is still a good idea.

derrabus:
Examples for fun times with callables:

* https://3v4l.org/1990k
* https://3v4l.org/cH1Ek

VincentLanglet:
> I appreciate the effort you've put in this PR and I'm honestly sorry that I've jumped onto this train that late. But I'm genuinely not convinced that this change is still a good idea.

I still think callable should still be deprecated in php before refusing support for them.
But, it's you library, your call.

The issue were opened for 2 years, I recommend to close the issue if the change won't never be accepted.

derrabus:
> I still think callable should still be deprecated in php before refusing support for them.

All right, since you're just repeating arguments you've already made, let's agree to disagree. This debate does not lead anywhere.

> The issue were opened for 2 years,

It might have been valid two years ago, but that does not mean we shouldn't reevaluate it after such a long time.

> I recommend to close the issue if the change won't never be accepted.

Yes, we should do that.

WDYT @greg0ire?

greg0ire:
Let's close this, I think @derrabus is making good point, and you're right, we should close the issue as well.

PR #322, opened 2022-08-24
Extract read-only part of Collection

### LINKED ISSUE #298: Adding a ReadableCollection interface
Currently, the only interface we have is Collection, which has both getter methods and mutator methods. Because of that, its generic type is invariant.
In many cases, we might want to write code that only cares about reading from a collection, which could then benefit from having a covariant interface instead. This can be solved by adding a ReadableCollection interface as a parent of Collection, moving all reading methods to the parent interface. That parent interface would use `@template-covariant`. This way, code that only cares about reading from a collection could use the ReadableCollection type, where passing a `ReadableCollection<Child>` to a method expecting `ReadableCollection<Parent>` would be accepted. There would be no BC break for users (the Collection interface itself would still have the same signature).
What do you think about that ? I can work on a PR if there is an interest.

greg0ire:
This allows us to mark the extracted interface as covariant, thus
allowing consumers to ensure through static analysis that a collection
will not be altered in a way that would no longer make it a collection
of items of a specific type.

Closes #298

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -32,6 +32,11 @@ explicitly retrieve an iterator though ``getIterator()`` which can then be
 used to iterate over the collection. You can not rely on the internal iterator
 of the collection being at a certain position unless you explicitly positioned it before.
 
+Methods that do not alter the collection or have template types
+appearing in invariant or contravariant positions are not directly
+defined in ``Doctrine\Common\Collections\Collection``, but are inherited
+from the ``Doctrine\Common\Collections\ReadableCollection`` interface.
+
 The methods available on the interface are:
 
 add
--- a/lib/Doctrine/Common/Collections/ArrayCollection.php
+++ b/lib/Doctrine/Common/Collections/ArrayCollection.php
@@ -231,6 +231,8 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @template TMaybeContained
      */
     public function contains($element)
     {
@@ -253,6 +255,12 @@ class ArrayCollection implements Collection, Selectable
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-param TMaybeContained $element
+     *
+     * @psalm-return (TMaybeContained is T ? TKey|false : false)
+     *
+     * @template TMaybeContained
      */
     public function indexOf($element)
     {
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -4,8 +4,6 @@ namespace Doctrine\Common\Collections;
 
 use ArrayAccess;
 use Closure;
-use Countable;
-use IteratorAggregate;
 
 /**
  * The missing (SPL) Collection/Array/OrderedMap interface.
@@ -26,10 +24,10 @@ use IteratorAggregate;
  *
  * @psalm-template TKey of array-key
  * @psalm-template T
- * @template-extends IteratorAggregate<TKey, T>
+ * @template-extends ReadableCollection<TKey, T>
  * @template-extends ArrayAccess<TKey, T>
  */
-interface Collection extends Countable, IteratorAggregate, ArrayAccess
+interface Collection extends ReadableCollection, ArrayAccess
 {
     /**
      * Adds an element at the end of the collection.
@@ -47,24 +45,6 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * @return void
      */
     public function clear();
-
-    /**
-     * Checks whether an element is contained in the collection.
-     * This is an O(n) operation, where n is the size of the collection.
-     *
-     * @param mixed $element The element to search for.
-     * @psalm-param T $element
-     *
-     * @return bool TRUE if the collection contains the element, FALSE otherwise.
-     */
-    public function contains($element);
-
-    /**
-     * Checks whether the collection is empty (contains no elements).
-     *
-     * @return bool TRUE if the collection is empty, FALSE otherwise.
-     */
-    public function isEmpty();
 
     /**
      * Removes the element at the specified index from the collection.
@@ -88,46 +68,6 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
     public function removeElement($element);
 
     /**
-     * Checks whether the collection contains an element with the specified key/index.
-     *
-     * @param string|int $key The key/index to check for.
-     * @psalm-param TKey $key
-     *
-     * @return bool TRUE if the collection contains an element with the specified key/index,
-     *              FALSE otherwise.
-     */
-    public function containsKey($key);
-
-    /**
-     * Gets the element at the specified key/index.
-     *
-     * @param string|int $key The key/index of the element to retrieve.
-     * @psalm-param TKey $key
-     *
-     * @return mixed
-     * @psalm-return T|null
-     */
-    public function get($key);
-
-    /**
-     * Gets all keys/indices of the collection.
-     *
-     * @return int[]|string[] The keys/indices of the collection, in the order of the corresponding
-     *               elements in the collection.
-     * @psalm-return TKey[]
-     */
-    public function getKeys();
-
-    /**
-     * Gets all values of the collection.
-     *
-     * @return mixed[] The values of all elements in the collection, in the
-     *                 order they appear in the collection.
-     * @psalm-return list<T>
-     */
-    public function getValues();
-
-    /**
      * Sets an element in the collection at the specified key/index.
      *
      * @param string|int $key   The key/index of the element to set.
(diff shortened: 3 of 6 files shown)


lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -48,24 +45,6 @@ public function add($element);
      */
     public function clear();
```
greg0ire, line 47: Should this one be moved up as well? Calling it still preserves the guarantee that the type of the items stays the same.

lib/Doctrine/Common/Collections/Collection.php
```diff
@@ -138,139 +77,4 @@ public function getValues();
...
-    /**
-     * Gets a native PHP array representation of the collection.
-     *
-     * @return mixed[]
-     * @psalm-return array<TKey,T>
-     */
-    public function toArray();
-
-    /**
-     * Sets the internal iterator to the first element in the collection and returns this element.
```
greg0ire, line 151: Is this fine? Since we are touching the internal iterator, I have a doubt.

stof: I would say that the internal iterator is fine (otherwise, you would also need to exclude `next()`

lib/Doctrine/Common/Collections/ReadableCollection.php
```diff
@@ -0,0 +1,209 @@
...
+
+namespace Doctrine\Common\Collections;
+
+use Closure;
+use Countable;
+use IteratorAggregate;
+
+/**
+ * @psalm-template TKey of array-key
+ * @psalm-template T
```
stof: I don't think we need this one, as we have `@template-covariant T`

lib/Doctrine/Common/Collections/ReadableCollection.php
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * @return bool TRUE, if the predicate yields TRUE for all elements, FALSE otherwise.
+     */
+    public function forAll(Closure $p);
+
+    /**
+     * Applies the given function to each element in the collection and returns
+     * a new collection with the elements returned by the function.
+     *
+     * @psalm-param Closure(T=):U $func
```
stof: why `=` ?

docs/en/index.rst
```diff
@@ -32,6 +32,17 @@ explicitly retrieve an iterator though ``getIterator()`` which can then be
...
+Methods that do not alter the collection are not directly defined in
+``Doctrine\Common\Collections\Collection``, but are inherited from the
+``Doctrine\Common\Collections\ReadableCollection`` interface.
+
+.. note::
+
+   When creating a method that only ever reads from the collection and
+   does not modify it, using the ``ReadableCollection`` interface should
+   be preferred, as it allows guaranteeing that the type of the
+   collection items will stay the same.
```
stof: I don't understand this supposed benefit. A `Collection<Dog>` does not allow changing the item type either.

The difference is that `ReadableCollection<Dog>` is a subtype of `ReadableCollection<Animal>` while `Collection<Dog>` is not a subtype of `Collection<Animal>`, making it a lot easier to implement a method that accepts a `ReadableCollection<MyInterface>` for instance when an implementation might have a `Collection<MyImplementation>`

lib/Doctrine/Common/Collections/Collection.php
stof: this one would not really fit the definition of `Readable` though. So I would keep it here even though it is not a blocker for the covariance of the template type.

lib/Doctrine/Common/Collections/ReadableCollection.php
greg0ire: Interesting, when I drop it, I get this:

```
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   Doctrine/Common/Collections/ReadableCollection.php
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  25     Template type T is declared as covariant, but occurs in contravariant position in parameter element of method Doctrine\Common\Collections\ReadableCollection::contains().
  142    Template type T is declared as covariant, but occurs in invariant position in return type of method Doctrine\Common\Collections\ReadableCollection::filter().
  179    Template type T is declared as covariant, but occurs in invariant position in return type of method Doctrine\Common\Collections\ReadableCollection::partition().
  179    Template type T is declared as covariant, but occurs in invariant position in return type of method Doctrine\Common\Collections\ReadableCollection::partition().
  192    Template type T is declared as covariant, but occurs in contravariant position in parameter element of method Doctrine\Common\Collections\ReadableCollection::indexOf().
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------



 [ERROR] Found 5 errors


╭─~/dev/doctrine-collections on readable-collection ✘ (fork/readable-collection)
╰$ vendor/bin/psalm
Target PHP version: 7.1 (inferred from composer.json)
Scanning files...
Analyzing files...

░░░E░░░

ERROR: InvalidTemplateParam - lib/Doctrine/Common/Collections/ReadableCollection.php:21:21 - Template param T of Doctrine\Common\Collections\ReadableCollection is marked covariant and cannot be used here (see https://psalm.dev/183)
     * @psalm-param T $element


ERROR: InvalidTemplateParam - lib/Doctrine/Common/Collections/ReadableCollection.php:126:21 - Template param T of Doctrine\Common\Collections\ReadableCollection is marked covariant and cannot be used here (see https://psalm.dev/183)
     * @psalm-param Closure(TKey=, T=):bool $p


ERROR: InvalidTemplateParam - lib/Doctrine/Common/Collections/ReadableCollection.php:137:21 - Template param T of Doctrine\Common\Collections\ReadableCollection is marked covariant and cannot be used here (see https://psalm.dev/183)
     * @psalm-param Closure(T=):bool $p


ERROR: InvalidTemplateParam - lib/Doctrine/Common/Collections/ReadableCollection.php:140:22 - Template param T of Doctrine\Common\Collections\ReadableCollection is marked covariant and cannot be used here (see https://psalm.dev/183)
     * @psalm-return Collection<TKey, T>


ERROR: InvalidTemplateParam - lib/Doctrine/Common/Collections/ReadableCollection.php:148:21 - Template param T of Doctrine\Common\Collections\ReadableCollection is marked covariant and cannot be used here (see https://psalm.dev/183)
     * @psalm-param Closure(TKey=, T=):bool $p


ERROR: InvalidTemplateParam - lib/Doctrine/Common/Collections/ReadableCollection.php:158:21 - Template param T of Doctrine\Common\Collections\ReadableCollection is marked covariant and cannot be used here (see https://psalm.dev/183)


ERROR: InvalidTemplateParam - lib/Doctrine/Common/Collections/ReadableCollection.php:172:21 - Template param T of Doctrine\Common\Collections\ReadableCollection is marked covariant and cannot be used here (see https://psalm.dev/183)
     * @psalm-param Closure(TKey=, T=):bool $p


ERROR: InvalidTemplateParam - lib/Doctrine/Common/Collections/ReadableCollection.php:177:22 - Template param T of Doctrine\Common\Collections\ReadableCollection is marked covariant and cannot be used here (see https://psalm.dev/183)
     * @psalm-return array{0: Collection<TKey, T>, 1: Collection<TKey, T>}


ERROR: InvalidTemplateParam - lib/Doctrine/Common/Collections/ReadableCollection.php:187:21 - Template param T of Doctrine\Common\Collections\ReadableCollection is marked covariant and cannot be used here (see https://psalm.dev/183)
     * @psalm-param T $element


9 errors found
3 other issues found.
You can display them with --show-info=true


```

lib/Doctrine/Common/Collections/ReadableCollection.php
greg0ire: That was introduced in 0c8bdd12640ee68286baf61fb5cae37b90c3af1a

@derrabus , do you remember why?

greg0ire: This was introduced in 744afc199626c1638113da4cd4bf2222c91dcd13

lib/Doctrine/Common/Collections/ReadableCollection.php
stof: hmm, `contains` and `indexOf` are indeed not contravariant on the value type (as we pass them as argument). So we might need to keep those out of that interface.

lib/Doctrine/Common/Collections/ReadableCollection.php
stof: `Closure(T=):U` means that the closure could be called with no argument, which then forces all consumers of the interface to write their callbacks with optional arguments. This does not make any sense.

On the other hand, given that PHP does not forbid passing extra arguments to a callback, it is totally fine to pass a callback expecting no argument to `Closure(T):U` (it will simply ignore the provided argument).

So `Closure(T):U` is the right type here. We want our code calling `$func` to always pass 1 argument.

docs/en/index.rst
greg0ire: Oh right I misunderstood, you cannot even have a template type appear in a parameter position. I'll change the docs.

lib/Doctrine/Common/Collections/ReadableCollection.php
greg0ire: Ok, I'll contribute that to 1.7.x

lib/Doctrine/Common/Collections/ReadableCollection.php
greg0ire: I moved all methods complained about by either tool back down to `Collection`, but I cannot say I fully understand for now. Maybe the naming of the interface should even be affected by that.

stof: See https://phpstan.org/blog/whats-up-with-template-covariant for the explaination of covariance vs invariance for the generic type.

greg0ire: It helps, thanks!

lib/Doctrine/Common/Collections/ArrayCollection.php
```diff
@@ -253,6 +255,12 @@ public function exists(Closure $p)
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-param TMaybeContained $element
+     *
+     * @psalm-return (TMaybeContained is T ? TKey|false : false)
```
greg0ire, line 261: If I don't copy/paste this here, I get
```
ERROR: InvalidReturnType - lib/Doctrine/Common/Collections/ArrayCollection.php:261:21 - The declared return type '(TKey:Doctrine\Common\Collections\ReadableCollection as array-key)|false' for Doctrine\Common\Collections\ArrayCollection::indexOf is incorrect, got '(TKey:Doctrine\Common\Collections\ArrayCollection as array-key)|false' (see https://psalm.dev/011)
    public function indexOf($element)
```
from psalm, not sure why

lib/Doctrine/Common/Collections/ReadableCollection.php
```diff
@@ -0,0 +1,213 @@
...
+     * @psalm-return T|null
+     */
+    public function get($key);
+
+    /**
+     * Gets all keys/indices of the collection.
+     *
+     * @return int[]|string[] The keys/indices of the collection, in the order of the corresponding
+     *               elements in the collection.
+     * @psalm-return TKey[]
```
stof: should probably be `list<TKey>` here to be more precise

lib/Doctrine/Common/Collections/ArrayCollection.php
AndrolGenhald: I'd have to spend a bit more time looking to be sure, but I'm guessing that's probably https://github.com/vimeo/psalm/issues/7553. Docblock inheritance and templates don't play well with one another in Psalm, probably due to the way `TKey` on `ReadableCollection` is a different type from `TKey` on `ArrayCollection` and the docblock inheritance isn't automatically translating it based on the `@implements`/`@extends` on the class/interface declaration.

SenseException (commented):
There's still a failed build.

greg0ire:
:facepalm: how did I not see that? Thanks for pointing it out!

greg0ire:
Ah that would be because the build is 12 minutes old :thinking:

greg0ire:
Addressed in #327

SenseException:
> Ah that would be because the build is 12 minutes old thinking

Yeah, someone tested the PR locally, then suddenly run into this problem and tested it in this PR. I don't know who that was though :innocent:

SenseException (approved):
I'm not sure about the impact on custom classes that extend a Doctrine Collection, but a sample worked fine locally.

PR #323, opened 2022-08-24
Mark closure arguments as always-provided

greg0ire:
Forcing consumers to take this into account makes no sense, and it is
fine to force implementors to provide those arguments.

### DIFF
--- a/lib/Doctrine/Common/Collections/Collection.php
+++ b/lib/Doctrine/Common/Collections/Collection.php
@@ -191,7 +191,7 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * Tests for the existence of an element that satisfies the given predicate.
      *
      * @param Closure $p The predicate.
-     * @psalm-param Closure(TKey=, T=):bool $p
+     * @psalm-param Closure(TKey, T):bool $p
      *
      * @return bool TRUE if the predicate is TRUE for at least one element, FALSE otherwise.
      */
@@ -202,7 +202,7 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * The order of the elements is preserved.
      *
      * @param Closure $p The predicate used for filtering.
-     * @psalm-param Closure(T=):bool $p
+     * @psalm-param Closure(T):bool $p
      *
      * @return Collection<mixed> A collection with the results of the filter operation.
      * @psalm-return Collection<TKey, T>
@@ -213,7 +213,7 @@ interface Collection extends Countable, IteratorAggregate, ArrayAccess
      * Tests whether the given predicate p holds for all elements of this collection.
      *
      * @param Closure $p The predicate.
-     * @psalm-param Closure(TKey=, T=):bool $p
+     * @psalm-param Closure(TKey, T):bool $p
      *
      * @return bool TRUE, if the predicate yields TRUE for all elements, FALSE otherwise.
      */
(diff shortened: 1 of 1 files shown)


stof:
As said in my comment on the other PR, this is not even because ignoring the argument makes no sense (you can still do it after that change).
The actual issue is that forcing all consumers to accept that the Collection calls the callback without argument makes no sense.

stof (changes requested):
The same needs to be done for other arguments accepting a callback.

greg0ire:
Making all the arguments optional was based on https://github.com/doctrine/collections/pull/177#discussion_r265717675.

By consumer, you mean some function that has a `Collection` in its scope, right? Not implementors of the interface?

Because that equal sign does not seem to mean the argument might not be provided, but rather that it's fine not to provide it:

- https://psalm.dev/r/132c51013b
- https://phpstan.org/r/f452b483-4fdf-468c-bef9-22f5aebdeb60

I'm closing because I don't thing anything needs to be change here after all.

stof:
@greg0ire your analysis is wrong. See https://phpstan.org/r/80ee8665-0ca3-4a29-aeba-31296dfe6409 which proves that it is also fine to pass a callback that has no argument without the `=`.
And see https://phpstan.org/r/3dc9bc1d-f8f9-4554-a0cc-603c34d8256d for a broken code allowed when using `=` (too bad that phpstan does not enforce the provided callback to make the argument optional during type checking, which is a bug in phpstan or psalm to me as https://psalm.dev/r/1b710fb4db is broken PHP code), while without `=`, it properly detects the wrong call to the callable: https://phpstan.org/r/5dd44dab-b3e4-4aa9-b38e-e963786ca194
And if psalm and phpstan fix their bug to properly detect the broken code, the `=` would force everyone to pass map callbacks that have optional arguments.

Btw, for a working implementation of checking callback signatures, try to write the same in Typescript (which does not have the equivalent bug) because the theory is the same (allowing usages that will work at runtime while catching invalid signature usages).

greg0ire:
@weirdan hi! Do you remember why you suggested making all the arguments optional in the first place: https://github.com/doctrine/collections/pull/177#discussion_r265717675 ? Was the situation maybe different at that time?

stof (dismissed):

stof:
@greg0ire you'll need to update #322 to sync it.

greg0ire:
I think I should retarget to 1.8.x now that I have realised that there is a small BC-break: implementors are now forced to provide those arguments.

stof:
@greg0ire all Doctrine implementors are providing the arguments. And I expect consumers of doctrine/collections to expect them to be provided too (too bad that SA tools don't report the wrong callbacks when the callback expects the argument but the method takes an argument.

But I'm find doing that in 1.8.x only.

stof:
However, note that if SA tools fix their detection of wrong callbacks passed to `callable(T=)`, the 1.7.x branch will become annoying for all *consumers* of doctrine/collections (rather than for implementors)

greg0ire:
Hmmm you're right. I think both scenarios are unlikely, but the one you last mentioned is the most likely. Retargeting back to 1.7.x then.

stof:
The second scenario is a **lot more** likely. phpstan already fixed the bug in the past, and they reverted the fix due to this wrong phpdoc in doctrine/collections: https://github.com/phpstan/phpstan/issues/7320

ghost:
> I think I should retarget to 1.8.x now that I have realised that there is a small BC-break: implementors are now forced to provide those arguments.

@greg0ire If back then you realised that this is a small BC-break then why were you targeting 1.8.x? BC breaks are supposed to target major versions, aren't they? Does this lib follow semantic versioning?

greg0ire:
It tries to. Is the bc break that big?

stof:
@javaDeveloperKid this is a small BC break for implementors of the interface (assuming that they were not already passing the arguments all the time). But for *users* of the interface, this is actually a bugfix, because almost none of them were making the arguments of their listeners optional (which was what the old signature was meaning, and phpstan fixed their bug not reporting it), especially given that all Doctrine-provided implementation have always passed those arguments all the time.

PR #331, opened 2022-09-27
Postpone return type additions to 3.0

greg0ire:
They prevent writing compatibility layers that would work with version 1 and 2 easily.

Prompted by https://github.com/doctrine/orm/pull/10070

It might be interesting to review the new diff with the stable branch: https://github.com/doctrine/collections/compare/1.7.x...greg0ire:collections:revert-return-type-additions


greg0ire:
If you feel some types should be made final or internal instead, please let me know.

derrabus (approved):

PR #333, opened 2022-09-29
Update branch metadata

greg0ire:
This is supposed to be the last 1.x minor branch, hence why 1.9.x is not created.


derrabus:
Shall we also set `"maintained": false` for 1.7 and remove `"current": true` from that branch?

PR #335, opened 2022-09-29
Update branch metadata

greg0ire:
1.8.x is no longer marked as current since this branch is only supposed
  to receive bugfixes.

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -5,16 +5,26 @@
     "docsSlug": "doctrine-collections",
     "versions": [
         {
-            "name": "2.0",
-            "branchName": "2.0.x",
+            "name": "3.0",
+            "branchName": "3.0.x",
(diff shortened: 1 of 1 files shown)


.doctrine-project.json
```diff
@@ -5,16 +5,26 @@
...
+            "branchName": "2.0.x",
+            "slug": "stable",
+            "current": true
+        },
         {
             "name": "1.8",
             "branchName": "1.8.x",
-            "slug": "1.8",
-            "current": true
+            "slug": "1.8"
```
SenseException, line 27: Is 1.8 still maintained?

greg0ire: Yes, I would say so, until there are reasonably little downloads. For that we have to add support for collections 2 to ORM and ODM.

SenseException (approved):

PR #345, opened 2022-10-12
fix: use @psalm-template annotation to avoid clashes

redjam13:
`[Semantical Error] The annotation "@template" in method Doctrine\Common\Collections\ArrayCollection::contains() was never imported. Did you maybe forget to add a "use" statement for this annotation?`

This is an issue since https://github.com/doctrine/collections/compare/1.7.3...1.8.0.

See https://github.com/doctrine/collections/pull/186. History repeating itself I think 🙂

### DIFF
--- a/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
+++ b/lib/Doctrine/Common/Collections/AbstractLazyCollection.php
@@ -62,7 +62,7 @@ abstract class AbstractLazyCollection implements Collection
     /**
      * {@inheritDoc}
      *
-     * @template TMaybeContained
+     * @psalm-template TMaybeContained
      */
     public function contains($element)
     {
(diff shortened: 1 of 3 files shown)


derrabus:
Shouldn't this be fixed by doctrine/annotations#298? What version of Doctrine Annotations are you using?

redjam13:
Thanks @derrabus. Yes, so it seems `doctrine/collections: 1.8.0` is a break if using `doctrine/annotations < 1.10.0`.

The project I am working on was maybe pegged at `doctrine/annotations: 1.8` perhaps due to the note on https://github.com/doctrine/annotations/releases/tag/1.9.0

Either way - we have inconsistent use of `@template` and `@pslam-template` here 🤷🏻‍♂️

derrabus:
Right, but I'd rather move forward and use the non-prefixed `@template` everywhere because that annotation has broad tooling support nowadays.

greg0ire:
Agree, this is a bug in `doctrine/annotations`, and versions below 1.13.x are no longer supported. Consider upgrading.

PR #348, opened 2022-11-09
`NOT` CompositeExpression in order to filter-out values

remi-san:
Some `Comparisons` do not provide the negation of the `Expression` (for example `EQ` has `NEQ`, but `CONTAINS` doesn't have `NCONTAINS`), making it difficult to create complex `Expressions` for use-cases like "I want to retrieve all books from the list where the title **contain** the word 'spider', but only if they **don't contain** the word 'spider-man'."

To do that, you'd have to filter the list with something like `Criteria::create()->where(Criteria::expr()->contains('title', 'Spider')`, store the result, and then filter it again with `Criteria::create()->where(Criteria::expr()->contains('title', 'Spider-man')`, and remove the elements retrieved by the second filter from the first filtered list.

With the  `NOT` expression, you would be able to do it with one simple operation:
```
Criteria::create()->where(
    Criteria::expr()->andX(
        Criteria::expr()->contains('title', 'Spider'),
        Criteria::expr()->not(
            Criteria::expr()->contains('title', 'Spider-man')
        )
    )
);
```

Of course, it would also allow a lot more of complexe cases where we would just want to inverse the result of an expression.

### DIFF
--- a/src/ExpressionBuilder.php
+++ b/src/ExpressionBuilder.php
@@ -30,6 +30,11 @@ class ExpressionBuilder
         return new CompositeExpression(CompositeExpression::TYPE_OR, $expressions);
     }
 
+    public function not(Expression $expression): CompositeExpression
+    {
+        return new CompositeExpression(CompositeExpression::TYPE_NOT, [$expression]);
+    }
+
     /** @return Comparison */
     public function eq(string $field, mixed $value)
     {
--- a/tests/Common/Collections/Expr/CompositeExpressionTest.php
+++ b/tests/Common/Collections/Expr/CompositeExpressionTest.php
@@ -14,21 +14,23 @@ use RuntimeException;
 /** @covers  \Doctrine\Common\Collections\Expr\CompositeExpression */
 class CompositeExpressionTest extends TestCase
 {
-    /** @return list<array{expression: mixed}> */
+    /** @return list<array{type:string, expressions: list<mixed>}> */
     public function invalidDataProvider(): array
     {
         return [
-            ['expression' => new Value('value')],
-            ['expression' => 'wrong-type'],
+            ['type' => CompositeExpression::TYPE_AND, 'expressions' => [new Value('value')]],
+            ['type' => CompositeExpression::TYPE_AND, 'expressions' => ['wrong-type']],
+            ['type' => CompositeExpression::TYPE_NOT, 'expressions' => [$this->createMock(Expression::class), $this->createMock(Expression::class)]],
         ];
     }
 
-    /** @dataProvider invalidDataProvider */
-    public function testExceptions(string|Value $expression): void
+    /**
+     * @param list<mixed> $expressions
+     *
+     * @dataProvider invalidDataProvider
+     */
+    public function testExceptions(string $type, array $expressions): void
     {
-        $type        = CompositeExpression::TYPE_AND;
-        $expressions = [$expression];
-
         $this->expectException(RuntimeException::class);
         new CompositeExpression($type, $expressions);
     }
--- a/docs/en/expression-builder.rst
+++ b/docs/en/expression-builder.rst
@@ -42,6 +42,18 @@ orX
 
     $collection->matching(new Criteria($expression));
 
+not
+---
+
+.. code-block:: php
+    $expressionBuilder = Criteria::expr();
+
+    $expression = $expressionBuilder->not(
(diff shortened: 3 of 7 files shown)


malarzm:
@remi-san thanks for the PR! Your changes on its own are looking good (except for PHPStan failure which may be unrelated). Few things that we (as Doctrine) should consider though:

1. ORM will need to support this in https://github.com/doctrine/orm/blob/2.13.x/lib/Doctrine/ORM/Query/QueryExpressionVisitor.php
2. ORM does not have a `NOT` operator so will need to map all non-complex operators to their negated counterparts (`>` -> `<=` etc)

FWIW ODM doesn't transform expressions to MongoDB queries.

greg0ire:
About PHPStan, see https://github.com/doctrine/collections/pull/349

malarzm:
Cool, one thing less ^^ So a rebase atop `2.0.x` will be fix CI, but we can do that later

tests/Common/Collections/Expr/CompositeExpressionTest.php
```diff
@@ -14,21 +14,23 @@
...
+            ['type' => CompositeExpression::TYPE_AND, 'expressions' => [new Value('value')]],
+            ['type' => CompositeExpression::TYPE_AND, 'expressions' => ['wrong-type']],
+            ['type' => CompositeExpression::TYPE_NOT, 'expressions' => [$this->createMock(Expression::class), $this->createMock(Expression::class)]],
         ];
     }
 
-    /** @dataProvider invalidDataProvider */
-    public function testExceptions(string|Value $expression): void
+    /**
+     * @param array{type:string, expressions: mixed} $expressions
```
greg0ire:
```suggestion
     * @param list<mixed> $expressions
```

tests/Common/Collections/Expr/CompositeExpressionTest.php
```diff
@@ -14,21 +14,23 @@
...
+     *
+     * @dataProvider invalidDataProvider
+     * */
```
greg0ire:
```suggestion
     */
```

tests/Common/Collections/Expr/CompositeExpressionTest.php
```diff
@@ -14,21 +14,23 @@
 /** @covers  \Doctrine\Common\Collections\Expr\CompositeExpression */
 class CompositeExpressionTest extends TestCase
 {
-    /** @return list<array{expression: mixed}> */
+    /** @return list<array{type:string, expressions: mixed}> */
```
greg0ire:
```suggestion
    /** @return list<array{type:string, expressions: list<mixed>}> */
```

remi-san:
Thank you for the feedback.

I rebased my branch so that `phpstan` can stop failing, and fixed the messed up phpdoc (thanks @greg0ire).

@malarzm I was expecting (fearing ? 😨) this kind of following task, I'll try submitting a PR for Doctrine ORM. Is the `NOT` operator not supported in the ORM on purpose or is it a good way to start introducing it?

src/ExpressionBuilder.php
```diff
@@ -30,6 +30,12 @@ public function orX(Expression ...$expressions)
         return new CompositeExpression(CompositeExpression::TYPE_OR, $expressions);
     }
 
+    /** @return CompositeExpression */
+    public function not(Expression $expression)
```
greg0ire:
```suggestion
    public function not(Expression $expression): CompositeExpression
```

remi-san:
@malarzm  I've created a PR on [`doctrine/orm`](https://github.com/doctrine/orm/pull/10234) as a proof the `NOT` operator can be handled quite easily (a lot more than what I expected).

malarzm:
Ah cool! I must have overlooked it while skimming through code yesterday :)

greg0ire:
@remi-san since this is a new feature, I've retargeted to 2.1.x.

The PR looks good to me, I think the docs need to be updated though, there should be at least something new in `docs/en/expressions.rst`, but please check other pages. Also, please squash your commits together (let me know if you need help on doing that).

remi-san:
@greg0ire Commits have been squashed and doc has been updated, tell me if there's anything else you'd like me to do.

greg0ire (approved):

malarzm (approved):

greg0ire:
Thanks @remi-san !

PR #349, opened 2022-11-14
Remove ignore rule for possibly null variable

greg0ire:
It seems PHPStan is now able to leverage the psalm-assert annotation that was added to exclude null from the possible types.


malarzm (approved):

malarzm:
Thanks @greg0ire!

PR #355, opened 2022-12-08
Use more precise return type for count method

VincentLanglet:
Countable is supposed to return
```
@return int<0, max>
```
(https://github.com/vimeo/psalm/pull/8861)

So I would say it's better to rely on inherit doc rather than overriding the return tag ?
@greg0ire

### DIFF
--- a/src/ArrayCollection.php
+++ b/src/ArrayCollection.php
@@ -296,7 +296,9 @@ class ArrayCollection implements Collection, Selectable, Stringable
         return array_values($this->elements);
     }
 
-    /** @return int */
+    /**
+     * {@inheritDoc}
+     */
     #[ReturnTypeWillChange]
     public function count()
     {
--- a/src/ArrayCollection.php
+++ b/src/ArrayCollection.php
@@ -298,6 +298,8 @@ class ArrayCollection implements Collection, Selectable, Stringable
 
     /**
      * {@inheritDoc}
+     *
+     * @return int<0, max>
      */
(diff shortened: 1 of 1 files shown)


greg0ire:
If it doesn't break what https://github.com/doctrine/collections/pull/353 solves, that looks good indeed.

VincentLanglet:
> If it doesn't break what #353 solves, that looks good indeed.

Yeah, I was wondering if I needed the `@return int<0, max>` for this...
Is there any test about it ?

Because the LazyCollection use inherit
https://github.com/doctrine/collections/blob/2.1.x/src/AbstractLazyCollection.php#L31-L35

src/ArrayCollection.php
```diff
@@ -296,7 +296,9 @@ public function getValues()
         return array_values($this->elements);
     }
 
-    /** @return int */
+    /**
+     * {@inheritDoc}
```
VincentLanglet, line 300:
```suggestion
     * {@inheritDoc}
     *
     * @return int<0, max>
```

VincentLanglet:
I just tried in my project @greg0ire and the `@return` is needed.
So I updated it to `@return int<0, max>` then. (I get no deprecation this way)

greg0ire (approved):
Thanks for triple checking, Vincent :pray:

SenseException (approved):

greg0ire:
Technically not a bugfix, but it's OK to target 2.1 IMO because the things it's improving is not released yet.

PR #366, opened 2023-02-27
Value Object comparison operations

psfpro:
Improvements for working with Value Object comporation. Value Object not necessarily being the same object for being equal.

> In [computer science](https://en.wikipedia.org/wiki/Computer_science), a value object is a small [object](https://en.wikipedia.org/wiki/Object_(computer_science)) that represents a simple [entity](https://en.wikipedia.org/wiki/Entity) whose equality is not based on [identity](https://en.wikipedia.org/wiki/Identity_(object-oriented_programming)): i.e. two value objects are equal when they have the same value, not necessarily being the same object.

https://en.wikipedia.org/wiki/Value_object

### DIFF
--- a/src/Expr/ClosureExpressionVisitor.php
+++ b/src/Expr/ClosureExpressionVisitor.php
@@ -127,8 +127,16 @@ class ClosureExpressionVisitor extends ExpressionVisitor
         $value = $comparison->getValue()->getValue();
 
         return match ($comparison->getOperator()) {
-            Comparison::EQ => static fn ($object): bool => self::getObjectFieldValue($object, $field) === $value,
-            Comparison::NEQ => static fn ($object): bool => self::getObjectFieldValue($object, $field) !== $value,
+            Comparison::EQ => static function ($object) use ($field, $value): bool {
+                $fieldValue = ClosureExpressionVisitor::getObjectFieldValue($object, $field);
+
+                return is_scalar($fieldValue) ? $fieldValue === $value : $fieldValue == $value;
+            },
+            Comparison::NEQ => static function ($object) use ($field, $value): bool {
+                $fieldValue = ClosureExpressionVisitor::getObjectFieldValue($object, $field);
+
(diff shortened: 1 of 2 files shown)


derrabus:
> Value Object not necessarily being the same object for being equal.

No, but if they're not the same object, we simply don't know how to compare them. I am open to making object comparison more extensible, but simply falling back to a non-strict comparison and crossing fingers doesn't seem like a good solution.

psfpro:
This is a documented way php object comparison

> When using the comparison operator (==), object variables are compared in a simple manner, namely: Two object instances are equal if they have the same attributes and values (values are compared with ==), and are instances of the same class.

https://www.php.net/manual/en/language.oop5.object-comparison.php

derrabus:
Thank you, I know how the `==` operator works.

psfpro:
We can go Java way object comparison
```php
Comparison::EQ => static function ($object) use ($field, $value): bool {
    $fieldValue = ClosureExpressionVisitor::getObjectFieldValue($object, $field);
    if (is_object($fieldValue) && method_exists($fieldValue, 'equals')) {
        return $fieldValue->equals($field);
    }

    return $fieldValue === $value;
},
```
or write cusom function

What do you think could be the best solution?

derrabus:
Duck-typing leads to a similar problem: We somewhat guess how equality in meant to be implemented for a given object.

vudaltsov:
I see 2 extension approaches here.

First is to add a `Comparable` interface on the Doctrine side. This will result in heavier coupling of userland classes to Doctrine and will not help with vendor classes like Ramsey UUID, for instance.

The second way is is to add comparators that have a `supports(string $class): bool` method. But they will need to be added statically, because there's no DI here...

So both solutions kinda suck...

derrabus:
Let's close the PR because this is not the solution we're looking for. If you still like to see this feature, please feel free to open an issue for further discussion.

PR #370, opened 2023-05-08
Upgrade to Psalm 5

greg0ire:


psalm.xml.dist
```diff
@@ -18,6 +20,18 @@
...
+            <errorLevel type="suppress">
+                <file name="src/AbstractLazyCollection.php"/>
+            </errorLevel>
+        </MixedInferredReturnType>
+
+        <MixedReturnStatement>
+            <errorLevel type="suppress">
+                <file name="src/AbstractLazyCollection.php"/>
+            </errorLevel>
+        </MixedReturnStatement>
```
greg0ire, line 33: Caused by

```
ERROR: MixedInferredReturnType - src/AbstractLazyCollection.php:277:21 - Could not verify return type 'TInitial|TReturn|null' for Doctrine\Common\Collections\AbstractLazyCollection::reduce (see https://psalm.dev/047)
    public function reduce(Closure $func, mixed $initial = null)


ERROR: MixedReturnStatement - src/AbstractLazyCollection.php:281:16 - Could not infer a return type (see https://psalm.dev/138)
        return $this->collection->reduce($func, $initial);
```

This seems to complex for Psalm.

derrabus (approved):

PR #373, opened 2023-07-06
Added the documentation.yml to the workflows

FlorianT03:
Test the github action that validates the documentation everytime it is changed

### DIFF
--- a/src/ArrayCollection.php
+++ b/src/ArrayCollection.php
@@ -429,6 +429,8 @@ class ArrayCollection implements Collection, Selectable, Stringable
     /**
      * Returns a string representation of this object.
      * {@inheritDoc}
+     *
+     * @return string
      */
     #[ReturnTypeWillChange]
     public function __toString()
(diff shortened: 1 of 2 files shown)


src/ArrayCollection.php
```diff
@@ -429,9 +429,11 @@ public function partition(Closure $p)
     /**
      * Returns a string representation of this object.
      * {@inheritDoc}
+     *
+     * @return string
      */
     #[ReturnTypeWillChange]
-    public function __toString()
+    public function __toString(): string
```
greg0ire: This needs to be removed:

```suggestion
    public function __toString()
```

We will eventually add that native return type declaration, but in the next major release.

src/ArrayCollection.php
```diff
@@ -444,6 +444,7 @@ public function clear()
         $this->elements = [];
     }
 
+
```
greg0ire: This change needs to be reverted
```suggestion
```

greg0ire (approved):

PR #379, opened 2023-09-28
Make returntypes consistent with implementation

### LINKED ISSUE #372: should the return type of ReadableCollection:filter() & ReadableCollection:map() not be static ?
hi,

This is more a question than an issue...  
Got the feeling the return types of `ReadableCollection:filter()` & `ReadableCollection:map()` should be `static` instead of the current `ReadableCollection` 

This is a simplified version of what I'm doing:

```php
class Foo
{
    private function __construct(
        public Collection $collection,
    )
    { }

    public static function of(Collection $collection): self
    {
        $collection = $collection->filter(fn($item) => $item->test());
        return new self($collection);
    }
}
``` 

And PhpStan complains about this.... because, as it is, `$collection->filter()` will return a `ReadableCollection` not a `Collection`, which is obviously what is to be expected IMHO.

By the way, when looking at the implementation of `ArrayCollection`, the defined return type is `static`.

But I might be completely wrong about this...  
Therefor asking the question before bluntly stating this is an issue...

quite possible the return type might need a change for other methods, did not take the time to review everything...  
would definitely submit a PR if the above question is indeed an issue

heiglandreas:
The implementation of filter and map are working on a Collection and not on a ReadableCollection. Therefore they will also return a Collection and not just a ReadableCollection.

Closes https://github.com/doctrine/collections/issues/372

### DIFF
--- a/src/AbstractLazyCollection.php
+++ b/src/AbstractLazyCollection.php
@@ -230,10 +230,7 @@ abstract class AbstractLazyCollection implements Collection
     }
 
     /**
-     * @psalm-param Closure(T, TKey):bool $p
-     *
-     * @return ReadableCollection<mixed>
-     * @psalm-return ReadableCollection<TKey, T>
+     * {@inheritDoc}
      */
     public function filter(Closure $p)
     {
@@ -254,13 +251,6 @@ abstract class AbstractLazyCollection implements Collection
 
     /**
      * {@inheritDoc}
-     *
-     * @psalm-param Closure(T):U $func
-     *
-     * @return ReadableCollection<mixed>
-     * @psalm-return ReadableCollection<TKey, U>
-     *
-     * @psalm-template U
      */
     public function map(Closure $func)
     {
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\Common\Collections;
 
 use ArrayAccess;
+use Closure;
 
 /**
  * The missing (SPL) Collection/Array/OrderedMap interface.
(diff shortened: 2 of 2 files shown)


greg0ire:
> The implementation of filter and map are working on a Collection and not on a ReadableCollection.

I think what both Psalm and PHPStan are saying here is that you don't actually have a guarantee of that according to the annotations on `Collection::filter` and `Collection::map`.

greg0ire:
I saw your toot about `fixup!` commits, and while I agree they're great, they're AFAIK not taken into account by Gitl{lab,hub}. Either you will have to rebase interactively, or we will have to squash merge. Let us know what your plan is regarding this.

heiglandreas:
Would be great to support that in Github et al, yeah!

But Inwanted to rebase that locally when I'm happy with it. Should have created a draft 🤦

derrabus:
I think, we're going to squash the PR in the end anyway.

greg0ire:
> Should have created a draft

You can still click "convert to draft" after the PR is created :wink:

greg0ire (approved):

greg0ire:
To be squash merged or squashed. When squashing, I think we should apply the same changes I just applied to the title and description (apart maybe from the one with about using a Github keyword, as it only matters to Github).

greg0ire:
Oh and I think the target branch is wrong, it should be 2.2 since this is not a bugfix

stof:
Note that this allows the behavior to match what we had in 1.7.0 (before extracting ReadableCollection)

greg0ire:
Ah yes… then forget about my comment, since it can be considered as a fix for a regression.

heiglandreas:
All rebased and squashed so that it now is only one commit

SenseException (approved):
Regression as in "not done on purpose in 2.0", even though the major version would allow such a change. But I don't expect much of a negative impact on this change/fix.

derrabus (approved):
LGTM.

Should we apply this change to the `partition()` method as well?

PR #380, opened 2023-09-29
Improve return type for Collection::partition

heiglandreas:
Currently when partitioning a Collection-object it will return an array of ReadableCollections unless that return-type is overwritten in the implementation.

This change makes sure that any implementation of a Collection actually returns an array of Collections instead.

This is a follow-up PR to #379 as suggested by @derrabus in https://github.com/doctrine/collections/pull/379#pullrequestreview-1649861837

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -100,4 +100,14 @@ interface Collection extends ReadableCollection, ArrayAccess
      * @psalm-return Collection<TKey, T>
      */
     public function filter(Closure $p);
+
+    /**
+     * {@inheritDoc}
+
+     * @return Collection<mixed>[] An array with two elements. The first element contains the collection
+     *                      of elements where the predicate returned TRUE, the second element
+     *                      contains the collection of elements where the predicate returned FALSE.
+     * @psalm-return array{0: Collection<TKey, T>, 1: Collection<TKey, T>}
+     */
+    public function partition(Closure $p);
 }


greg0ire:
There is a conflict, probably with what was merged in #379

src/Collection.php
```diff
@@ -79,4 +80,14 @@ public function removeElement(mixed $element);
      * @return void
      */
     public function set(string|int $key, mixed $value);
+
+    /**
+     * {@inheritDoc}
+
+     * @return Collection<mixed>[] An array with two elements. The first element contains the collection
```
stof, line 107: Using `Collection<mixed>` here does not make sense. Tools supporting generics should see `Collection<TKey, T>`.

heiglandreas: As this is a copy/paste from the docs of the ReadableCollection I'd then also fix that there...

And also fix the conflict

heiglandreas: After checking with the sources: So far the `@return` annotations all contain references to `mixed`. Should this then be changed everywhere?

Personally I'd use solely `@return array{0: Collection<TKey, T>, 1: Collection<TKey, T>}` and skip the `@return-psalm` annotation...

derrabus: I agree, these annotations are consistent with `ReadableCollection::partition()`. Merging `@return` and `@psalm-return` is probably out of scope here.

derrabus (approved):

PR #382, opened 2023-10-03
Disable UselessInheritDocComment sniff

greg0ire:
According to a maintainer of `slevomat/coding-standard`, it goes against other sniffs we use in this repository.

See https://github.com/doctrine/collections/pull/381#issuecomment-1745206747


SenseException:
Does it make sense to put the link to the comment above the excluded sniff for context?

derrabus (approved):

PR #385, opened 2023-12-30
improvement: count() now costs O(1) instad of O(n)

snipershady:
Now the method `count()` costs O(1) instad of O(n)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 /phpcs.xml
 /.phpcs-cache
 /.phpunit.result.cache
+/nbproject/private/
+/nbproject/
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,3 @@
 /phpcs.xml
 /.phpcs-cache
 /.phpunit.result.cache
-/nbproject/private/
-/nbproject/
--- a/src/ArrayCollection.php
+++ b/src/ArrayCollection.php
@@ -70,7 +70,7 @@ class ArrayCollection implements Collection, Selectable, Stringable
     public function __construct(array $elements = [])
     {
         $this->elements = $elements;
-        $this->size     = count($elements);
+        $this->size     = empty($elements) ? 0 : count($elements);
     }
 
     /**
--- a/src/ArrayCollection.php
+++ b/src/ArrayCollection.php
@@ -58,9 +58,10 @@ class ArrayCollection implements Collection, Selectable, Stringable
     
     /**
      * The number of elements in this colleciton
+     * 
      * @var int
      */
-    private int $size = 0;
+    private int $size;
 
     /**
      * Initializes a new ArrayCollection.
@@ -152,8 +153,11 @@ class ArrayCollection implements Collection, Selectable, Stringable
         }
 
         $removed = $this->elements[$key];
+        
         unset($this->elements[$key]);
+        
         $this->size--;
+        
         return $removed;
     }
 
@@ -169,7 +173,9 @@ class ArrayCollection implements Collection, Selectable, Stringable
         }
 
         unset($this->elements[$key]);
+        
         $this->size--;
+        
         return true;
     }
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -4,3 +4,5 @@
 /phpcs.xml
 /.phpcs-cache
 /.phpunit.result.cache
+/nbproject/private/
+/nbproject/
```
derrabus: Please revert.

src/ArrayCollection.php
```diff
@@ -65,6 +71,7 @@ class ArrayCollection implements Collection, Selectable, Stringable
     public function __construct(array $elements = [])
     {
         $this->elements = $elements;
+        $this->size = count($elements);
```
derrabus: This increases the cost of the constructor call from O(1) to O(n).

derrabus (changes requested):
Does this change actually solve a problem for you? Did you run any benchmarks?

The obvious blocker for me is the fact that the constructor call gets unnecessarily expensive with your change.

Also since the logic behind the `count()` method gets more complex with your change, we will probably need more tests around the calculation.

That being said, I'm not sure the added complexity is worth it, tbh.

.gitignore
snipershady: done

src/ArrayCollection.php
snipershady: only if you create a collection from another collection.
All new collection costs O(1)

.gitignore
greg0ire: Pro tip: [use a global gitignore](https://greg0ire.fr/git-gud/#/23)

snipershady:
The Static Analysis problem, from PHPstan, wasn't a problem of this PR
I hadn't modified that part of the doc

greg0ire:
> The Static Analysis problem, from PHPstan, wasn't a problem of this PR

Are you willing to bet? :stuck_out_tongue: There's an easy way you could make sure of that, you know :wink:

src/ArrayCollection.php
```diff
@@ -98,7 +104,7 @@ public function first()
      * @psalm-template K of array-key
      * @psalm-template V
      */
-    protected function createFrom(array $elements)
+    protected function createFrom(array $elements): static
```
greg0ire, line 107: Also, the "changes" view makes it obvious which part of your changes is causing the issue.

src/ArrayCollection.php
derrabus: > only if you create a collection from another collection.

Which is a totally valid thing to do! This is a blocker!

src/ArrayCollection.php
```diff
@@ -65,6 +70,7 @@ class ArrayCollection implements Collection, Selectable, Stringable
     public function __construct(array $elements = [])
     {
         $this->elements = $elements;
+        $this->size     = empty($elements) ? 0 : count($elements);
```
derrabus, line 73: You don't need to avoid `count()` _on an empty array_. That's a cheap operation. `count`ing non-empty arrays preemptively is what troubles me.

src/ArrayCollection.php
derrabus, line 107: Adding a native return type is a breaking change. We must not do that in a minor or bugfix release. Apart from that, this change seems unrelated to the goal of this PR.

.gitignore
```diff
@@ -3,4 +3,4 @@
 /build
 /phpcs.xml
 /.phpcs-cache
-/.phpunit.result.cache
+/.phpunit.result.cache
```
derrabus, line 6: Please revert. 🙃

derrabus (changes requested):

malarzm:
FWIW `count` is O(1) when used on arrays, PHP tracks number of elements in a hash table on its own https://github.com/php/php-src/blob/ea6110f52e40c2b2112364df6d6ca99cb81ab63f/Zend/zend_types.h#L407

PR #387, opened 2024-02-16
Avoid risky truthy-falsy comparison

greg0ire:
Stumbled upon this while working on https://github.com/doctrine/collections/pull/386

### DIFF
--- a/src/ArrayCollection.php
+++ b/src/ArrayCollection.php
@@ -480,7 +480,7 @@ class ArrayCollection implements Collection, Selectable, Stringable
         $offset = $criteria->getFirstResult();
         $length = $criteria->getMaxResults();
 
-        if ($offset || $length) {
+        if ($offset !== null && $offset > 0 || $length !== null && $length > 0) {
             $filtered = array_slice($filtered, (int) $offset, $length, true);
         }
 


src/ArrayCollection.php
```diff
@@ -480,7 +480,7 @@ public function matching(Criteria $criteria)
         $offset = $criteria->getFirstResult();
         $length = $criteria->getMaxResults();
 
-        if ($offset || $length) {
+        if ($offset !== null && $offset > 0 || $length !== null && $length > 0) {
             $filtered = array_slice($filtered, (int) $offset, $length, true);
```
SenseException, line 484:
```suggestion
        if ($offset !== null && $offset > 0 || $length !== null && $length > 0) {
            $filtered = array_slice($filtered, $offset, $length, true);
```

SenseException (changes requested):
Is this PR a cosmetic one?

greg0ire:
> Is this PR a cosmetic one?

Sorry, I should have mentioned that it fixes the build by addressing an issue found by Psalm

greg0ire:
@SenseException after applying your suggestion the tests break like this:

```
1) Doctrine\Tests\Common\Collections\ArrayCollectionTest::testMatchingWithSlicingPreserveKeys with data set "preserve keys on maxresult only" (array(1, 2, 3, 4), array(1, 2), null, 2)
TypeError: array_slice(): Argument #2 ($offset) must be of type int, null given

/home/greg/dev/doctrine-collections/src/ArrayCollection.php:484
/home/greg/dev/doctrine-collections/tests/Common/Collections/BaseArrayCollectionTest.php:356
```

SenseException (approved):
Because of the `||` it can be null. I forgot that case is still valid.

PR #389, opened 2024-02-17
Introduce Order enum

greg0ire:
The plan is to ultimately use when dealing with `Criteria::$orderings`

### DIFF
--- a/src/Criteria.php
+++ b/src/Criteria.php
@@ -19,12 +19,15 @@ use function strtoupper;
  */
 class Criteria
 {
-    final public const ASC  = 'ASC';
+    /** @deprecated use Order::Ascending instead */
+    final public const ASC = 'ASC';
+
+    /** @deprecated use Order::Descending instead */
     final public const DESC = 'DESC';
 
     private static ExpressionBuilder|null $expressionBuilder = null;
 
-    /** @var array<string, string> */
+    /** @var array<string, Order> */
     private array $orderings = [];
 
     private int|null $firstResult = null;
@@ -57,7 +60,7 @@ class Criteria
     /**
      * Construct a new Criteria.
      *
-     * @param array<string, string>|null $orderings
+     * @param array<string, string|Order>|null $orderings
      */
     public function __construct(
         private Expression|null $expression = null,
@@ -151,9 +154,32 @@ class Criteria
     /**
      * Gets the current orderings of this Criteria.
      *
+     * @deprecated use orderings() instead
+     *
      * @return array<string, string>
      */
     public function getOrderings()
+    {
+        Deprecation::trigger(
+            'doctrine/collections',
+            'https://github.com/doctrine/collections/pull/389',
+            'Calling %s() is deprecated. Use %s::orderings() instead.',
+            __METHOD__,
+            self::class,
+        );
+
+        return array_map(
+            static fn (Order $ordering): string => $ordering->value,
+            $this->orderings,
+        );
+    }
+
+    /**
+     * Gets the current orderings of this Criteria.
+     *
+     * @return array<string, Order>
+     */
+    public function orderings(): array
     {
         return $this->orderings;
     }
@@ -161,19 +187,38 @@ class Criteria
     /**
      * Sets the ordering of the result of this Criteria.
      *
-     * Keys are field and values are the order, being either ASC or DESC.
+     * Keys are field and values are the order, being a valid Order enum case.
      *
-     * @see Criteria::ASC
-     * @see Criteria::DESC
+     * @see Order::Ascending
+     * @see Order::Descending
      *
-     * @param array<string, string> $orderings
+     * @param array<string, string|Order> $orderings
      *
      * @return $this
      */
     public function orderBy(array $orderings)
     {
         $this->orderings = array_map(
-            static fn (string $ordering): string => strtoupper($ordering) === self::ASC ? self::ASC : self::DESC,
+            static function (string|Order $ordering): Order {
+                if ($ordering instanceof Order) {
+                    return $ordering;
+                }
+
+                static $triggered = false;
+
+                if (! $triggered) {
+                    Deprecation::trigger(
+                        'doctrine/collections',
+                        'https://github.com/doctrine/collections/pull/389',
+                        'Passing non-Order enum values to %s() is deprecated. Pass Order enum values instead.',
+                        __METHOD__,
+                    );
+                }
+
+                $triggered = true;
+
+                return strtoupper($ordering) === Order::Ascending->value ? Order::Ascending : Order::Descending;
+            },
             $orderings,
         );
 
--- a/src/Order.php
+++ b/src/Order.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Common\Collections;
+
+enum Order: string
+{
+    case Ascending  = 'ASC';
+    case Descending = 'DESC';
+}
--- a/tests/BaseArrayCollectionTest.php
+++ b/tests/BaseArrayCollectionTest.php
@@ -6,6 +6,7 @@ namespace Doctrine\Tests\Common\Collections;
 
 use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Collections\Criteria;
+use Doctrine\Common\Collections\Order;
 use Doctrine\Common\Collections\Selectable;
 use PHPUnit\Framework\TestCase;
 use stdClass;
@@ -334,6 +335,15 @@ abstract class BaseArrayCollectionTest extends TestCase
                 ->matching(new Criteria(null, ['sortField' => Criteria::ASC]))
                 ->toArray(),
         );
+        self::assertSame(
+            [
+                'object2' => $object2,
+                'object1' => $object1,
+            ],
+            $collection
+                ->matching(new Criteria(null, ['sortField' => Order::Ascending]))
+                ->toArray(),
+        );
     }
 
     /**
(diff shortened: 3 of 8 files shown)


ThomasLandauer:
Yeah, looks good to me :-)
If you merge it, I will take care of the rest of the docs, as promised in https://github.com/doctrine/orm/pull/11263#issue-2137531540

src/Criteria.php
```diff
@@ -155,16 +155,19 @@ public function getWhereExpression()
...
      * Sets the ordering of the result of this Criteria.
      *
      * Keys are field and values are the order, being either ASC or DESC.
      *
-     * @see Criteria::ASC
-     * @see Criteria::DESC
+     * @see Order::ASC
+     * @see Order::DESC
      *
      * @param array<string, string> $orderings
```
greg0ire: In the future, we would have `array<string, Order>` here.

src/Criteria.php
```diff
@@ -155,16 +155,19 @@ public function getWhereExpression()
      */
```
greg0ire, line 160: In the future, we would have `@return array<string, Order>` here

greg0ire: Maybe I should use `array<string, string|Order>` right now so that people start adapting their code, even if in practice we don't break BC and stay with `array<string, string>` ? :thinking: either that or introducing a new method.

derrabus: We should provide a way to opt into the future behavior of this method. We could do so by adding a boolean flag (meh) or by adding a new method and deprecating this one.

src/Order.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Common\Collections;
+
+enum Order: string
+{
+    case ASC  = 'ASC';
+    case DESC = 'DESC';
```
derrabus:
```suggestion
    case Ascending  = 'ASC';
    case Descending = 'DESC';
```

WDYT?

src/Criteria.php
greg0ire: Ok, I'll try to find a new name.

src/Criteria.php
```diff
@@ -155,25 +155,34 @@ public function getWhereExpression()
...
      *
      * Keys are field and values are the order, being either ASC or DESC.
      *
-     * @see Criteria::ASC
-     * @see Criteria::DESC
+     * @see Order::ASC
+     * @see Order::DESC
      *
-     * @param array<string, string> $orderings
+     * @param array<string, string|Order> $orderings
```
derrabus, line 195: The types of the constructor signature have to be adjusted as well.

src/Criteria.php
```diff
@@ -155,25 +155,34 @@ public function getWhereExpression()
...
     public function orderBy(array $orderings)
     {
         $this->orderings = array_map(
-            static fn (string $ordering): string => strtoupper($ordering) === self::ASC ? self::ASC : self::DESC,
+            static function (string|Order $ordering): Order {
+                if ($ordering instanceof Order) {
+                    return $ordering;
+                }
+
+                return strtoupper($ordering) === Order::ASC->value ? Order::ASC : Order::DESC;
```
derrabus: Trigger a deprecation?

src/Criteria.php
```diff
@@ -155,25 +155,34 @@ public function getWhereExpression()
...
+        );
     }
 
     /**
      * Sets the ordering of the result of this Criteria.
      *
      * Keys are field and values are the order, being either ASC or DESC.
      *
-     * @see Criteria::ASC
-     * @see Criteria::DESC
```
derrabus, line 167: Deprecate the constants?

tests/BaseArrayCollectionTest.php
```diff
@@ -331,7 +332,7 @@ public function testMatchingWithSortingPreserveKeys(): void
                 'object1' => $object1,
             ],
             $collection
-                ->matching(new Criteria(null, ['sortField' => Criteria::ASC]))
+                ->matching(new Criteria(null, ['sortField' => Order::ASC]))
```
derrabus: By simply changing the tests, you lose coverage for the old way of using this API. Codecov rightfully complains about that.

derrabus (changes requested):

src/Criteria.php
greg0ire: Since there is no setter, I went with just `orderings()`

src/Criteria.php
greg0ire: Should we give the ORM a chance to upgrade first?

ThomasLandauer:
I guess this will then also be the way to order collections?:
```php
#[ORM\OneToMany(...)
#[ORM\OrderBy(value: ['foo' => Order::Ascending])
private Collection $bar;
```

greg0ire:
Yes 👍

src/Criteria.php
derrabus: I don't think we need to wait for that.

greg0ire: Ok, let me add that deprecation layer then.

src/Criteria.php
greg0ire: done

src/Criteria.php
greg0ire: done

greg0ire:
Ping @derrabus :pray:

derrabus (approved):

ThomasLandauer:
Which branch/version of the docs should I edit?
* DBAL 4.1 https://www.doctrine-project.org/projects/doctrine-dbal/en/4.1/index.html
* ORM 3.1 https://www.doctrine-project.org/projects/doctrine-orm/en/3.1/index.html

greg0ire:
I think it depends, we have to look at the code, and see if the new names are passed as is or if some kind of translation/method signature changes are required.

BTW, why are you mentioning DBAL? It does not depend on collections, does it?

ThomasLandauer:
I'm talking about the docs, not the code!
DBAL docs have some occurrences, e.g. https://www.doctrine-project.org/projects/doctrine-dbal/en/4.0/reference/query-builder.html#order-by-clause

So should I replace all 'ASC' with the new `Order::Ascending` everywhere in ORM 3.1 (assuming that by that time everybody will have upgraded their `collections`)? Or add a note "New in version xy" like at https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/tutorials/extra-lazy-associations.html#extra-lazy-associations ? - This could become a mess, if there are many occurrences (which I guess there are...)

greg0ire:
> I'm talking about the docs, not the code!

I know, and please don't shout!

>  DBAL docs have some occurrences, e.g. https://www.doctrine-project.org/projects/doctrine-dbal/en/4.0/reference/query-builder.html#order-by-clause

That's an SQL query builder. This has nothing to do with collections whatsoever. `doctrine/dbal` should work without `doctrine/collections` installed. If you modify those docs to refer to `Order`, you will generate support tickets.

> So should I replace all 'ASC' with the new Order::Ascending everywhere in ORM 3.1 (assuming that by that time everybody will have upgraded their collections)? Or add a note "New in version xy" like at https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/tutorials/extra-lazy-associations.html#extra-lazy-associations ? - This could become a mess, if there are many occurrences (which I guess there are...)

Maybe, wait a bit, we have to evaluate which ones are going to be OK, and which ones require a breaking change. For instance consider this:

```php
class ClassFromTheORM
{
    public function methodFromTheOrm(string $ordering)
    {
        Criteria::orderBy(['field' => $ordering);
    }
}
```

Can't recommend `Order::Ascending` for that one.

ThomasLandauer:
> and please don't shout!

Sorry :-)

> That's an SQL query builder. This has nothing to do with collections whatsoever.

I thought this is what this PR addresses?! - Moving from `->orderBy('username', 'ASC')` to `->orderBy('username', Order::Ascending)`?
So what is deprecated then?

greg0ire:
>  So what is deprecated then?

What you just said + referring to `Criteria::ASC`. We can see that it happens in the ORM, here for instance: https://github.com/doctrine/orm/blob/2a8802af122921ad9c40a266a74ce8c210c18e44/src/PersistentCollection.php#L588 or here: https://github.com/doctrine/orm/blob/2a8802af122921ad9c40a266a74ce8c210c18e44/src/Mapping/Driver/XmlDriver.php#L408

I don't think you will be able to find similar stuff in the DBAL.

ThomasLandauer:
OK, gotcha.
* https://www.doctrine-project.org/projects/doctrine-dbal/en/4.0/reference/query-builder.html#order-by-clause is about the **SQL** Query Builder (not DQL Query Builder as I thought)
* Indeed there are no other occurrences of ASC or DESC in the DBAL docs.
* In the ORM docs there are fewer occurrences than I thought, most of them are on https://www.doctrine-project.org/projects/doctrine-orm/en/3.0/reference/query-builder.html So should I start with that page now or wait?

greg0ire:
You should IMO wait until we at least:
- release `doctrine/collections` 2.2.0
- contribute something that allows using `Order::Ascending` (right now, `QueryBuilder::orderBy()` accepts a string, and `Order::Ascending` is no string, `Order::Ascending->value` is).

jifer:
We just upgraded doctrine packages. Changing `Criteria` constants to enum was straightforward. We do love enums.

Problem is that we used this everywhere where QueryBuilder is involved (a lot), and it was quite disappointing seeing that this is not going to work.

We go from this:
```
$this->getEntityManager()
            ->getRepository(SomeEntity::class)
            ->createQueryBuilder('h')
            ->orderBy('h.id', Criteria::DESC)
            ->getQuery()
            ->getResult();
```
to this:

```
$this->getEntityManager()
            ->getRepository(SomeEntity::class)
            ->createQueryBuilder('h')
            ->orderBy('h.id', Order::Descending->value)
            ->getQuery()
            ->getResult();
```

Yes, I do understand that Collections & ORM are separate, but ORM relays on Collections as dependency.
So can we count on:
a) Allowing `Order` as `orderBy` second param
b) or dedicated enum for QB that will replace string.

derrabus:
Neither a) nor b) is planned at the moment. That being said, this is the wrong repository for this discussion.

PR #393, opened 2024-02-18
Upgrade to PHPUnit 10

greg0ire:


tests/BaseArrayCollectionTestCase.php
```diff
@@ -20,7 +20,7 @@
 use function next;
 use function reset;
 
-abstract class BaseArrayCollectionTest extends TestCase
+abstract class BaseArrayCollectionTestCase extends TestCase
```
derrabus:
```suggestion
abstract class ArrayCollectionTestCase extends TestCase
```

The `Base` prefix feels redundant now.

tests/Expr/CompositeExpressionTest.php
```diff
@@ -15,12 +15,26 @@
...
+                        {
+                        }
+                    },
+                    new class () implements Expression {
+                        public function visit(ExpressionVisitor $visitor)
+                        {
+                        }
+                    },
+                ],
+            ],
```
derrabus: You could use `self::createStub()` to replace `$this->createMock()` in a static method.

greg0ire: TIL `createMock` is not statick, but `createStub` is :bulb:

derrabus (approved):

PR #394, opened 2024-02-21
reduce method: more accurate types

gharlan:


src/ReadableCollection.php
```diff
@@ -229,11 +229,11 @@ public function findFirst(Closure $p);
      * Applies iteratively the given function to each element in the collection,
      * so as to reduce the collection to a single value.
      *
-     * @psalm-param Closure(TReturn|TInitial|null, T):(TInitial|TReturn) $func
-     * @psalm-param TInitial|null $initial
+     * @psalm-param Closure(TReturn|TInitial, T):TReturn $func
+     * @psalm-param TInitial $initial
```
gharlan, line 233: If `$initial` is not `null`, the closure does not need to accept `null` value for first param.

greg0ire:
This was introduced in #252 
@VincentLanglet please review :pray:

VincentLanglet:
> This was introduced in #252 @VincentLanglet please review 🙏

I dunno where I found the types, but i copied them from somewhere else,
like https://github.com/vimeo/psalm/issues/5574 for example.

Searching for similar types in psalm, I found
https://github.com/vimeo/psalm/blob/c401490265497fd9d025459b5d3dbe29d80394fe/stubs/extensions/ds.phpstub#L196-L203
https://github.com/phpstan/phpstan-src/blob/0f9fda8461421672e5b2f90807289d9207796c7f/stubs/arrayFunctions.stub#L3-L13

This PR seems to go in the right direction

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @gharlan !

PR #408, opened 2024-04-03
Add isNotNull to ExpressionBuilder

raziel057:
Currently we have the ``isNull``, but ``isNotNull`` is missing.

I propose to add it.

### DIFF
--- a/docs/en/expression-builder.rst
+++ b/docs/en/expression-builder.rst
@@ -124,6 +124,16 @@ isNull
 
     $collection->matching(new Criteria($expression));
 
+isNotNull
+------
+
+.. code-block:: php
+    $expressionBuilder = Criteria::expr();
+
+    $expression = $expressionBuilder->isNotNull('foo');
+
+    $collection->matching(new Criteria($expression));
+
 in
 ---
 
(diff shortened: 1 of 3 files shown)


src/ExpressionBuilder.php
```diff
@@ -77,6 +77,12 @@ public function isNull(string $field)
         return new Comparison($field, Comparison::EQ, new Value(null));
     }
 
+    /** @return Comparison */
+    public function isNotNull(string $field)
```
derrabus:
```suggestion
    public function isNotNull(string $field): Comparison
```

derrabus (changes requested):
Thank you.

src/ExpressionBuilder.php
raziel057: I was not sure whether it would be preferable to include the return type in the DockBloc for consistency with other methods or to add it as a native type as the method is a new one.

docs/en/expression-builder.rst
```diff
@@ -124,6 +124,16 @@ isNull
 
     $collection->matching(new Criteria($expression));
 
+isNotNull
+------
```
greg0ire:
```suggestion
---------
```

derrabus (approved):

PR #414, opened 2024-05-03
Enable dependabot

greg0ire:
Let us automate bumping workflow actions


derrabus:
Can we move that file to our .github repository?

SenseException (approved):
> Can we move that file to our .github repository?

@derrabus [It doesn't look like it](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#configuration-options-for-the-dependabotyml-file).

PR #418, opened 2024-05-03
Set label of GH actions dependabot PRs to CI

greg0ire:
People could get confused when seeing such changes under "dependencies" in our CHANGELOG.


greg0ire:
If you are OK with this I will relabel past PRs with CI.

SenseException (approved):

PR #420, opened 2024-05-18
Add `prepend()` to Collection interface

ker0x:
This is a new attempt to introduce a `prepend()` method to the `Collection` interface, after #162.


greg0ire:
Sorry, but we cannot accept additions to that interface without the contributor making a strong case for it, because of the S and I of [SOLID](https://en.wikipedia.org/wiki/SOLID). There are already ~ 30 methods on collection objects, please let use know why this is needed according to you.

derrabus:
Closing due to lack of feedback.

PR #421, opened 2024-06-16
Throw exception if matching() is called on a Collection with scalar items

priyadi:
Previously, it would result in an obscure `TypeError`: 

> Doctrine\Common\Collections\Expr\ClosureExpressionVisitor::getObjectFieldValue(): Argument #1 ($object) must be of type object|array, int given, called in /home/priyadi/Projects/rekalogika/collections/src/Expr/ClosureExpressionVisitor.php on line 113


malarzm:
Sorry but your change does not make sense: you're widening type just to throw another exception that's saying the same thing as current TypeError. Also the method you're changing is public so it can be called in any context, not just from `matching` method.

PR #422, opened 2024-07-05
Use types declared in doc-block as return type hints

pl-github:
I noticed that [intersection types are used as return types in ORM](https://github.com/doctrine/orm/blob/1fe1a6a048dd420d06704f72b296a463237d7603/src/EntityRepository.php#L199), but not in the Collections library.

I have adjusted the two places affected here.


SenseException:
I'd say that this also needs to be added as an information to the [UPGRADE.md](https://github.com/doctrine/collections/blob/3.0.x/UPGRADE.md#signature-changes) file. It's a deviation to what is described in 2.0.

PR #424, opened 2024-08-27
Leverage new array functions

derrabus:
PHP 8.4 will add a couple of array functions that we can make use of:

https://wiki.php.net/rfc/array_find

I'm using Symfony's polyfill to make sure the functions are always available. If we don't want to add that dependency, we could also work with a PHP version check inside the methods I've changed.

### DIFF
--- a/stubs/array_find.phpstub
+++ b/stubs/array_find.phpstub
@@ -1,0 +1,12 @@
+<?php
+
+/**
+ * @param array<TKey, TValue>                      $array
+ * @param callable(TValue $value, TKey $key): bool $callback
+ *
+ * @return TValue|null
+ *
+ * @template TKey of array-key
+ * @template TValue
+ */
+function array_find(array $array, callable $callback): mixed {}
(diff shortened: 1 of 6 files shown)


stubs/array_find.phpstub
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * @param array<TKey, TValue>                      $array
+ * @param callable(TValue $value, TKey $key): bool $callback
+ *
+ * @return TValue|null
+ *
+ * @template TKey of array-key
+ * @template TValue
+ */
+function array_find(array $array, callable $callback): mixed {}
```
greg0ire, line 12: Why is a stub needed for this function but not the others?

greg0ire (dismissed):
I'm OK with the polyfill… it's an extra dependency, but tiny, likely to be already installed, and not here to stay.

stubs/array_find.phpstub
derrabus: Psalm does not know about the new functions yet. We need this stub, so Psalm can infer the correct return type for `array_find()`. The other two new functions I've used always return a boolean, so the native return types might be just good enough.

greg0ire (approved):

PR #438, opened 2024-12-10
Setup website schema validation workflow

greg0ire:
This allows to validate changes made against .doctrine-project.json
against the corresponding JSON schema.


greg0ire:
The build passes with a valid file, and fails with an invalid file 👍

SenseException (approved):

PR #444, opened 2025-02-27
Fix typo

OskarStark:


greg0ire:
I'm not sure it's a typo, see line 29: https://github.com/doctrine/collections/blob/a9dc31b5e5120320039ddfbf612032277dea5b5b/UPGRADE.md?plain=1#L29

OskarStark:
Agreed, but then we either need 3 columns or relax the 3.0 info to the 3.0 upgrade guide.
I mean do I really need this info when upgrading to 2.x then? 🤔

greg0ire:
>  I mean do I really need this info when upgrading to 2.x then? 🤔

_need_, maybe not, but it should be interesting to you, because you might want to change your signatures to be forward-compatible with 3.x, and not just compatible with 2.x

> then we either need 3 columns 

not _needed_, but could clarify the situation

> relax the 3.0 info to the 3.0 upgrade guide.

Sorry, what do you mean?

OskarStark:
> Sorry, what do you mean?

I mean, if not needed, we can keep 2.x here without the return type and add 2.x to 3.x table to the UPGRADE guide for 3.0

greg0ire:
OK, but IMO, it's not just about what's needed, it's more about what will ensure you do as few migrations as possible. Knowing what the return types are going to be can help if you have types that are `final` and you can afford to become compatible with 2.x _and_ 3.x at once.

OskarStark:
Agree, let's close here. Thanks for your feedback

Issue #437, opened 2024-12-10
[Collections] Criteria::orderBy() ignores all but first ordering in multi-column sort


tautis154:
### Bug Report

|    Q                                        |   A
|-------------------------------------------- | ------
| Version                                     | 2.1.2

#### Summary
The Criteria::orderBy() method in Doctrine Collections only applies the first ordering pair from the provided array, completely ignoring subsequent orderings. This prevents proper multi-column sorting in Collection filtering.

#### Current behavior
When using Criteria::orderBy() with multiple sort criteria, only the first key/value pair in the array is applied to the sorting. Any additional orderings are ignored, and changing their direction (ASC/DESC) has no effect on the results.
For example, with two records having the same validFrom date:

When using `Criteria::orderBy()` with multiple sort criteria, only the first key/value pair in the array is applied to the sorting. Any additional orderings are ignored, and changing their direction (ASC/DESC) has no effect on the results.

For example, with two records having the same `validFrom` date:
```php
// Records:
// id=5, validFrom='2024-12-10 16:04:11'
// id=6, validFrom='2024-12-10 16:04:11'

$collection->matching(
    Criteria::create()
        ->orderBy([
            'validFrom' => Criteria::DESC,
            'id' => Criteria::DESC  // This is ignored
        ])
);

// Returns id=5 first, despite having same validFrom and lower id
```

Even changing `'id' => Criteria::ASC` has no effect on the results, demonstrating that the second ordering is completely ignored.

#### Expected behavior
All orderings in the array should be applied in sequence, matching SQL ORDER BY behavior:
- Primary sorting should be by the first criteria
- When values are equal for the first ordering, subsequent orderings should be used as tiebreakers
- In the example above, id=6 should come first when sorting by validFrom DESC, id DESC

#### How to reproduce
```php
use Doctrine\Common\Collections\Criteria;
use Doctrine\Common\Collections\ArrayCollection;

class PriceList
{
    private int $id;
    private \DateTimeImmutable $validFrom;
    
    public function __construct(int $id, \DateTimeImmutable $validFrom) 
    {
        $this->id = $id;
        $this->validFrom = $validFrom;
    }

    public function getId(): int
    {
        return $this->id;
    }

    public function getValidFrom(): \DateTimeImmutable
    {
        return $this->validFrom;
    }
}

// Test setup
$list1 = new PriceList(5, new \DateTimeImmutable('2024-12-10 16:04:11'));
[...]

Issue #361, opened 2017-09-20
ClosureExpressionVisitor::sortByField() fails when comparing Objects/Dates


t11n:

beberlei:
The ExpressionBuilder has this class level comment about only "scalar" values being supported for interoperability. Obviously a comment is well hidden and hard to spot. We could throw an exception if values are not a scalar value to enforce this, for one big whopping BC break. But given the amount of issues about this it might be worthwhile.

mpdude:
I may be wrong, but I believe this issue belongs to the doctrine/collections repo?

/cc @greg0ire

mpdude:
This caught me today and cost quite some time to figure out...

It's especially surprising in the ORM when you have a collection of entities and want to order by multiple fields where one contains a `DateTime[Immutable]` values:

```php
return $this->collection
   ->matching(
      Criteria::create()
         ->where(...something...)
         ->orderBy(['some_date_field' => Criteria::DESC, 'other_field' => Criteria::DESC])
   );
```

When the collection is not yet initialized, the query will be executed in the database where the `DateTime` is a scalar values (e.g. `2025-03-25`), and `other_field` is sorted as one would expect.

When the collection has been initialized (is an `ArrayCollection`), the fields will be distinct `DateTime[Immutable]` object instances. This means the `$next` closure will not be evaluated – effectively limiting the `orderBy` up to the first `DateTime[Immutable]` field given, but stopping there and not considering any remaining sort conditions.

mpdude:
Potential fix at #454

Issue #136, opened 2017-11-27
ClosureExpressionVisitor::sortByField() fails when comparing Objects/Dates


t11n:
In method `ClosureExpressionVisitor::sortByField()` we have the following sort function...

```
    return function ($a, $b) use ($name, $next, $orientation) {
        $aValue = ClosureExpressionVisitor::getObjectFieldValue($a, $name);
        $bValue = ClosureExpressionVisitor::getObjectFieldValue($b, $name);

        if ($aValue === $bValue) {
            return $next($a, $b);
        }

        return (($aValue > $bValue) ? 1 : -1) * $orientation;
    };
```

... which does not work with Objects, especially DateTime objects. I do not know if it makes sense to compare objects in general if they are greater or lesser, but it definitely makes sense for DateTime objects.

The problem is, that equal DateTime objects will not be recognised by `$aValue === $bValue`. Instead `$aValue > $bValue` will evaluate to false in these cases. Not a big deal you may think, as the a element is put below the b element, which is fine, as they are equal. That's right, but $next will neither be executed.

In my use case, sorting by date is the first order criteria, sorting by ID is the second. That code will never evaluate the second criteria, though.

Possible fix:
```
        if ($aValue instanceof \DateTime && $aValue == $bValue || $aValue === $bValue) {
            return $next($a, $b);
        }
```

You have to judge whether it makes sense to compare other object types. For DateTimes however, this is necessary, I suppose. Sorting by date with a second order by criteria is not possible otherwise.

Please have a look into the matter.

mpdude:
#437 is about the same issue. Also see #361 and #454.

PR #454, opened 2025-03-25
Allow to treat `DateTimeInterface` instances as scalar values when comparing or sorting

mpdude:
The `ClosureExpressionVisitor::walkComparison()` method uses strict comparison checks for the `EQ` and `NEQ` operations. Since 4b1d1b57445b28362744448e8ded24474ba00e26, the `IN` and `NIN` checks _don't_ use strict checks when objects are concerned. Less/greater [equal] than checks don't even have the concept of strict checking in PHP.

Similarly, `ClosureExpressionVisitor::sortByField()` implements multi-field sorting in a way that considers the "next" sort field only when there is _strict_ equality on the values of the previous field.

This can be very surprising when using the collection filtering API (i. e. `Selectable`) on collections in Doctrine ORM to filter or sort fields that are represented by `DateTime` or `DateTimeImmutable`.

As long as the collection has not been initialized, a database-level filter or sort is applied, which effectively makes date/time data behave as a value (or scalar).

Once the collection has been initialized, in-memory comparisons will give different results. Depending on how new entities are created, they might actually use the _same_ DateTime instances. Once they have been hydrated by the ORM, most probably they never do.

This means collection filtering for date/time values does not work predictably for `EQ`/`NEQ`, whilst the other comparisons behave as one would expect.

Sorting through the API will effectively stop multi-field sort orderings at the first date/time field involved (reported as #361).

I think we should consider instances of `DateTimeInterface` as values, not applying strict comparisons to them.

Of course, doing so would be a BC break with regard to filtering and sorting. Thus, this PR adds a new `Criteria::treatDateTimeAsScalar()` method that flips `Criteria` instances into the new mode of operation. That flag is passed on later to `ClosureExpressionVisitor` where the actual comparisons take place.

We could discuss whether not using the new mode should raise a deprecation warning. For now, however, I think we could go ahead with just adding it as a feature flag.

#### Related

Other issues about values, especially `DateTime` not being comparable as one would expect:

* #136 
* #361 
* #437
* #68

### DIFF
--- a/src/Criteria.php
+++ b/src/Criteria.php
@@ -30,6 +30,8 @@ class Criteria
     /** @var array<string, Order> */
     private array $orderings = [];
 
+    private bool $treatDateTimeAsScalar = false;
+
     private int|null $firstResult = null;
     private int|null $maxResults  = null;
 
@@ -224,6 +226,18 @@ class Criteria
         );
 
         return $this;
+    }
+
+    public function treatDateTimeAsScalar(): self
+    {
+        $this->treatDateTimeAsScalar = true;
+
+        return $this;
+    }
+
+    public function getTreatDateTimeAsScalar(): bool
+    {
+        return $this->treatDateTimeAsScalar;
     }
 
     /**
--- a/src/Expr/ClosureExpressionVisitor.php
+++ b/src/Expr/ClosureExpressionVisitor.php
@@ -6,11 +6,13 @@ namespace Doctrine\Common\Collections\Expr;
 
 use ArrayAccess;
 use Closure;
+use DateTimeInterface;
 use RuntimeException;
 
 use function array_all;
 use function array_any;
 use function explode;
+use function func_get_args;
 use function in_array;
 use function is_array;
 use function is_scalar;
@@ -31,6 +33,10 @@ use function strtoupper;
  */
 class ClosureExpressionVisitor extends ExpressionVisitor
 {
+    public function __construct(private readonly bool $treatDateTimeAsScalar = false)
+    {
+    }
+
     /**
      * Accesses the field of a given object. This field has to be public
      * directly or indirectly (through an accessor get*, is*, or a magic
@@ -103,16 +109,22 @@ class ClosureExpressionVisitor extends ExpressionVisitor
      */
     public static function sortByField(string $name, int $orientation = 1, Closure|null $next = null)
     {
+        if (isset(func_get_args()[3])) {
+            $orderDateTimeAsScalar = (bool) func_get_args()[3];
+        } else {
+            $orderDateTimeAsScalar = false;
+        }
+
         if (! $next) {
             $next = static fn (): int => 0;
         }
 
-        return static function ($a, $b) use ($name, $next, $orientation): int {
+        return static function ($a, $b) use ($name, $next, $orientation, $orderDateTimeAsScalar): int {
             $aValue = ClosureExpressionVisitor::getObjectFieldValue($a, $name);
 
             $bValue = ClosureExpressionVisitor::getObjectFieldValue($b, $name);
 
-            if ($aValue === $bValue) {
+            if ($aValue === $bValue || ($orderDateTimeAsScalar && $aValue instanceof DateTimeInterface && $bValue instanceof DateTimeInterface && $aValue == $bValue)) {
                 return $next($a, $b);
             }
 
@@ -129,8 +141,24 @@ class ClosureExpressionVisitor extends ExpressionVisitor
         $value = $comparison->getValue()->getValue();
 
         return match ($comparison->getOperator()) {
-            Comparison::EQ => static fn ($object): bool => self::getObjectFieldValue($object, $field) === $value,
-            Comparison::NEQ => static fn ($object): bool => self::getObjectFieldValue($object, $field) !== $value,
+            Comparison::EQ => function ($object) use ($field, $value): bool {
+                $fieldValue = self::getObjectFieldValue($object, $field);
+
+                if ($this->treatDateTimeAsScalar && $fieldValue instanceof DateTimeInterface) {
+                    return $fieldValue == $value;
+                }
+
+                return $fieldValue === $value;
+            },
+            Comparison::NEQ => function ($object) use ($field, $value): bool {
+                $fieldValue = self::getObjectFieldValue($object, $field);
+
+                if ($this->treatDateTimeAsScalar && $fieldValue instanceof DateTimeInterface) {
+                    return $fieldValue != $value;
+                }
+
+                return $fieldValue !== $value;
+            },
             Comparison::LT => static fn ($object): bool => self::getObjectFieldValue($object, $field) < $value,
             Comparison::LTE => static fn ($object): bool => self::getObjectFieldValue($object, $field) <= $value,
             Comparison::GT => static fn ($object): bool => self::getObjectFieldValue($object, $field) > $value,
--- a/src/Expr/ClosureExpressionVisitor.php
+++ b/src/Expr/ClosureExpressionVisitor.php
@@ -109,11 +109,7 @@ class ClosureExpressionVisitor extends ExpressionVisitor
      */
     public static function sortByField(string $name, int $orientation = 1, Closure|null $next = null)
     {
-        if (isset(func_get_args()[3])) {
-            $orderDateTimeAsScalar = (bool) func_get_args()[3];
-        } else {
-            $orderDateTimeAsScalar = false;
-        }
+        $orderDateTimeAsScalar = (bool) (func_get_args()[3] ?? false);
 
         if (! $next) {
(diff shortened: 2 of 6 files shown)


mpdude:
How to tell PHPCS that I do not want to make strict comparison as it suggests?

greg0ire:
You can run `vendor/bin/phpcs -s` to find which rule is causing this, then you can ignore it in `phpcs.xml.dist`.

greg0ire:
If that is a new feature, then you should target 2.4.x

src/Criteria.php
```diff
@@ -226,6 +228,20 @@ static function (string|Order $ordering) use ($method): Order {
         return $this;
     }
 
+    /** @return $this */
+    public function treatDateTimeAsScalar()
```
greg0ire:
```suggestion
    public function treatDateTimeAsScalar(): self
```

src/Criteria.php
```diff
@@ -226,6 +228,20 @@ static function (string|Order $ordering) use ($method): Order {
...
+    {
+        $this->treatDateTimeAsScalar = true;
+
+        return $this;
+    }
+
+    /** @return bool */
+    public function getTreatDateTimeAsScalar()
```
greg0ire:
```suggestion
    public function getTreatDateTimeAsScalar(): bool
```

src/Expr/ClosureExpressionVisitor.php
```diff
@@ -101,18 +106,18 @@ public static function getObjectFieldValue(object|array $object, string $field)
      *
      * @return Closure
      */
-    public static function sortByField(string $name, int $orientation = 1, Closure|null $next = null)
+    public static function sortByField(string $name, int $orientation = 1, Closure|null $next = null, bool $orderDateTimeAsScalar = false)
```
greg0ire: This is a breaking change for extending classes, isn't it? If yes, let's implement it with `func_get_arg()` instead?

tests/ArrayCollectionTestCase.php
```diff
@@ -457,4 +458,48 @@ public function testMultiColumnSortAppliesAllSorts(): void
...
+
+        $actual = $collection
+            ->matching((new Criteria(null, ['foo' => Order::Descending, 'bar' => Order::Ascending]))->treatDateTimeAsScalar())
+            ->toArray();
+
+        self::assertEquals($expected, $actual); // does not check array key order
+        self::assertSame(array_keys($expected), array_keys($actual));
+    }
+
+    public function testMatchingDateTimeEqualityAsScalar()
```
greg0ire:
```suggestion
    public function testMatchingDateTimeEqualityAsScalar(): void
```

src/Expr/ClosureExpressionVisitor.php
mpdude: Thank you for the suggestion, changed.

greg0ire:
Re-building with more PHP versions (my hope is that Codecov will stop wrongly complaining about some of your patch being uncovered).

greg0ire:
Nope, still not working :frowning_face:

tests/ArrayCollectionTestCase.php
```diff
@@ -457,4 +458,48 @@ public function testMultiColumnSortAppliesAllSorts(): void
...
+            2 => ['foo' => new DateTimeImmutable('2025-01-01 12:00:00'), 'bar' => new DateTimeImmutable('2025-01-03 12:00:00')],
+            1 => ['foo' => new DateTimeImmutable('2025-01-01 12:00:00'), 'bar' => new DateTimeImmutable('2025-01-04 12:00:00')],
+        ];
+
+        if (! $this->isSelectable($collection)) {
+            $this->markTestSkipped('Collection does not support Selectable interface');
+        }
+
+        $actual = $collection
+            ->matching((new Criteria(null, ['foo' => Order::Descending, 'bar' => Order::Ascending]))->treatDateTimeAsScalar())
```
greg0ire, line 483: I would expect this call to result in `treatDateTimeAsScalar` being covered. :thinking:

greg0ire: Oh, I guess this explains it: https://github.com/doctrine/collections/blob/793b4dfd3252dab123f5adb35e55474602ec4d23/tests/ArrayCollectionTest.php#L16

greg0ire (dismissed):

mpdude:
> Nope, still not working ☹️

Would it help if I rebased this?

greg0ire:
I don't think it would, because of https://github.com/doctrine/collections/pull/454#discussion_r2015782292
Also, I don't think its' a big deal if this getter is not covered, because it has zero complexity, and is still executed in tests.

src/Expr/ClosureExpressionVisitor.php
```diff
@@ -103,16 +109,22 @@ public static function getObjectFieldValue(object|array $object, string $field)
      */
     public static function sortByField(string $name, int $orientation = 1, Closure|null $next = null)
     {
+        if (isset(func_get_args()[3])) {
+            $orderDateTimeAsScalar = (bool) func_get_args()[3];
+        } else {
+            $orderDateTimeAsScalar = false;
+        }
```
SenseException:
```suggestion
        $orderDateTimeAsScalar = (bool) (func_get_args()[3] ?? false);
```

SenseException (dismissed):

mpdude:
Review approvals were dismissed due to merging of a suggestion

mpdude:
Maybe I should add this to the documentation as well, at least once after #456 has been merged

greg0ire (dismissed):

derrabus:
I don't feel comfortable with adding a very specific flag about `DateTime` objects to very generic class such as `Criteria` which otherwise wouldn't know anything about DateTime. Once we merge this, people might feel they need the same for `Number` objects, UUIDs or their own custom domain value object. Maybe we need something like https://github.com/sebastianbergmann/comparator to let people define equality. By the way, we do have similar issues in the ORM when it comes to using value objects as primary keys.

mpdude:
Definetly good point! I was wondering about that as well, but ultimately the `DateTime`s were the only value objects built into PHP that I was aware of.

To my understanding, everything in the ORM is by-value-ish anyway, since there is no such concept as identity-based checking at the database level.

What if the expression design allowed users to make the distinction between "equals" and "same as" as different operators?

Regarding the comparator library, keep in mind that we should not add anything here that we can not deliver or approximate in the ORM counterpart.

derrabus:
> Definetly good point! I was wondering about that as well, but ultimately the `DateTime`s were the only value objects built into PHP that I was aware of.

`Number` is another example. It's a very recent addition though.

> What if the expression design allowed users to make the distinction between "equals" and "same as" as different operators?

That might not be sufficient. Two objects might be considered equal even if not all properties match. It might even depend on the context: Should two dates be equal if they describe the same point in time, but in different timezones? What if I use `DateTime` objects to represent dates and the two objects have different milliseconds values?

> Regarding the comparator library, keep in mind that we should not add anything here that we can not deliver or approximate in the ORM counterpart.

I did not mean to suggest that we should use that library. I merely wanted to stress out that the problem of defining equality is complex enough so that people have created whole libraries around it.

A way forward here could be to replace your method `treatDateTimeAsScalar()` with something more generic like `setComparator()` which accepts an instance of our own `Comparator` interface.

```php
interface Comparator
{
    /**
     * @throws UncomparableException
     */
    public function equals(mixed $a, mixed $b): bool;
}
```

mpdude:
Maybe adding this feature on `Criteria` in doctrine/collections is not too difficult. On the other hand, how would you envision it to work at the ORM level in doctrine/orm, when having DBMS-based matching of collections?

I know, this should not be a primary concern from the collections point of view, but on the other hand, any such addition could make it more difficult to make the difference between in-memory and in-database matching on collection as transparent as possible.

derrabus:
> Maybe adding this feature on `Criteria` in doctrine/collections is not too difficult. On the other hand, how would you envision it to work at the ORM level in doctrine/orm, when having DBMS-based matching of collections?

You're right, my comparator approach won't work here. But to be fair, your flag wouldn't have any effect either as soon as we delegate determining equality to an external system such as the database.

mpdude:
The flag was meant as a BC way to make DateTime behave in memory as it behaves in ORM/database anyway.

mpdude:
We discussed this at lentgh during the 2025 Hackathon and agreed that we need a more general solution for all kinds of value objects.

 At the same time, we cannot maintain a list or knowledge about all the potential value objects and libraries for them that might be out there. The other way round, we cannot expect such libraries to implement dedicated Doctrine interfaces.

derrabus:
> The other way round, we cannot expect such libraries to implement dedicated Doctrine interfaces.

No, but we can expect someone who wants to integrate Collections and a specific value object to implement a piece of logic that tells Collections how to compare those.

Issue #460, opened 2025-03-10
Add merge method to ArrayCollection


dozsan:
### Feature Request
The ArrayCollection-on class should have a merge method

#### Why
When I want to add ManyToMany items, the doctrine deletes them first and then adds them, but this can be used to prevent it from deleting what we don't want to delete

#### How

```php
public function merge(Collection $collection): self
    {
        foreach ($this->getValues() as $element) {
            if (!$collection->contains($element)) {
                $this->removeElement($element);
            }
        }

        foreach ($collection as $element) {
            if (!$this->contains($element)) {
                $this->add($element);
            }
        }

        return $this;
    }
```

garak:
This should be opened on doctrine/collections instead

demotuulia:
I made my suggestion for this in the PR
[https://github.com/doctrine/collections/pull/462](https://github.com/doctrine/collections/pull/462)

derrabus:
The `ArrayCollection` is just the default implementation of the collection contracts. I don't believe that we need functionality to that class that is not supported by any other collection.

Also, would this merge function do anything beyond the following?

```php
return new ArrayCollection(array_merge(
    $collectionA->toArray(),
    $collectionB->toArray(),
));
```

If not, I'd like to close this issue.

derrabus:
Closing as explained.

PR #462, opened 2025-04-23
#460 Array Collection merge

demotuulia:
After the issue 
[https://github.com/doctrine/collections/issues/460](https://github.com/doctrine/collections/issues/460)
I made this solotution, which works like php array_merge.

See how it works 
find merge in
[https://github.com/demotuulia/doctrine_collections/blob/460/docs/en/index.rst](https://github.com/demotuulia/doctrine_collections/blob/460/docs/en/index.rst)
and in the test
[https://github.com/demotuulia/doctrine_collections/blob/eb577dc91f3cfd910b79e7c20d72630e37615a3c/tests/ArrayCollectionTest.php#L41](https://github.com/demotuulia/doctrine_collections/blob/eb577dc91f3cfd910b79e7c20d72630e37615a3c/tests/ArrayCollectionTest.php#L41)

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -311,6 +311,24 @@ Extracts a slice of $length elements starting at position $offset from the Colle
 
     $slice = $collection->slice(1, 2); // [1, 2]
 
+
+merge
+^^^^^
+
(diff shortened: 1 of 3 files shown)


derrabus (changes requested):
I don't think we need this feature, see https://github.com/doctrine/collections/issues/460#issuecomment-3194689033

PR #466, opened 2025-06-25
Adding flatMap

BenMakesGames:
I've run into use-cases where something like `flatMap`/`SelectMany` would be rad. for example, we currently resort to somersaults like this:

```php
return new ArrayCollection(
    array_merge(
        ...$this->someCollection->map(fn(Thing $o) => $o->getSomeOtherCollection())->toArray()
    )
);
```

wouldn't it be lovely to simply:

```php
return $this->someCollection->flatMap(fn(Thing $o) => $o->getSomeOtherCollection());
```

this PR adds an implementation of `flatMap` that matches the behavior of JavaScript, in that elements which AREN'T arrays are just kinda' included, for example, in JS:

`[ [1, 2, 3], 4, [5, 6, 7] ].flatMap(x => x)` yields `[ 1, 2, 3, 4, 5, 6, 7 ]`

personally I find this a little surprising - it feels a little too loosely-typed for my liking - but as PHP is ultimately a loosely-typed language, and Doctrine Collection methods are using the same names as the JS methods, and PHP & JS are often used in combination, I thought it would be appropriate to match the JS `flatMap` behavior. but I don't feel strongly about it, and am happy to change it.

I'm also happy to add `flat`, if it feels odd to add `flatMap` without `flat`.

### DIFF
--- a/src/ArrayCollection.php
+++ b/src/ArrayCollection.php
@@ -359,6 +359,32 @@ class ArrayCollection implements Collection, Selectable, Stringable
     }
 
     /**
+     * Applies the given function to each element in the collection and flattens the results.
+     * The function should return an array or single element for each element, which will be
+     * flattened into a single array.
+     *
+     * @phpstan-param Closure(T):array<U>|U $func
+     *
+     * @return static
+     * @phpstan-return static<int, U>
+     *
+     * @phpstan-template U
+     */
+    public function flatMap(Closure $func)
+    {
+        $result = [];
+        foreach ($this->elements as $element) {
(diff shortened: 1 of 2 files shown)


derrabus:
Thank you for your proposal. However, we usually don't add functionality to the `ArrayCollection` that we cannot provide on other collection implementations as well. Because we would need to change existing contracts or invent new ones, we're also very reluctant when it comes to adding the kind of utility function that you're proposing.

Because of this, I'm closing your PR. Thank you for understanding.

Issue #276, opened 2021-07-13
ClosureExpressionVisitor can not handle methods without the "get" prefix


gnumoksha:
Hello. I'm wondering why there is a regular expression check here:

https://github.com/doctrine/collections/blob/94918256daa6ac99c7e5774720c0e76f01936bda/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php#L55

I've methods that do not have "get" or "is" as a prefix, and they don't work because of that regex. Right now, I'm implementing the `__call` method on my object, but it would be nice if `ClosureExpressionVisitor` is able to find the methods.

malarzm:
This was proposed (https://github.com/doctrine/collections/pull/149) and declined before, also there's  https://github.com/doctrine/collections/pull/263 still open so I'm closing this as a duplicate.

Issue #170, opened 2018-10-31
DCOM-275: Collections\Expr\ClosureExpressionVisitor should support private properties as well as SQL visitors do.


doctrinebot:
Jira issue originally created by user anton.serdyuk:

Given I have some entity method like this:

```
    public function hasNewComments()
    {
        $newComments = $this->comments->matching(
            Criteria::create()
                ->where(Criteria::expr()->eq('new', true))
        );

        return count($newComments) > 0;
    }
```

Where $this->comments is OneToMany collection.

If $this->comments is EXTRA_LAZY, then this code will always produce SQL query to database and always will work fine. But if  it is not EXTRA_LAZY and appears to be loaded before (for eample count($this->comments) was called before), then this code will produce error "Trying to get private property Comment::$new from ClosureExpressionVisitor context".

I thins we should have something like reflection or \Closure::bind() stuff to get actual field value within ClosureExpressionVisitor regardless this field is private or public.

I know I can create getter getNew() or isNew(), but I believe this is known to be a good OOP practice to avoid unnecessary getters, and I am not sure collection api should make this decision for me.

janopae:
If a getter is available, should the ClosureExpressionVisitor prefer that getter, or still "force get" the private property?

I'd argue that the latter would be reasonable, as it would work transparently like how DQL queries and how lazy collections work. Also, it would be harder to break by just adding a getter. People might add getters that forbid accessing a property in a certain state, and that might lead to very edge case bugs that are hard to find.

However, not using getters for this any more would impose a BC break.

Any thoughts?

janopae:
Just found this comment by @Ocramius, which I think kind of agrees with my reasoning: https://github.com/doctrine/collections/pull/149#issuecomment-475982285

mpdude:
@beberlei Do you have any thoughts about this?

I am a bit confused since the code to be changed is in doctrine/collections, and from the point of view of that pacakge it might be considered a BC break to also inspect private properties.

From the ORM user's point of view, using a DQL expression to filter on `comment.new` (like in the initial example) checks the plain database value for uninitialized collections.

When the collection is initialized, it not only requires to have a public property or a getter/isser. Also the result might be different when that getter/isser does not simply return the plain value.

So, would it make sense to consider plain property values in that case, possibly using reflection to access also private values?

mpdude:
DDC-2838 is https://github.com/doctrine/orm/issues/3591

PR #472, opened 2025-10-08
Transition to using reflection-based, direct field access only

mpdude:
#### Motivation

There are a few issues about differences in behaviour when using the collection filtering API (the `Selectable` interface)  against collections that are database-backed (in ORM, these are `PersistentCollections`) vs. memory-based collections using `ArrayCollection` from this package here. 

For example:

* https://github.com/doctrine/orm/pull/11160
* #170
* https://github.com/doctrine/orm/issues/3591
* Maybe https://github.com/doctrine/orm/issues/11021

Database-based matching can work on the raw field values only, as those values are persisted to the database and there is no PHP code involved when filtering at the database level.

Memory-based matching currently tries a [series of access methods on the objects](https://www.doctrine-project.org/projects/doctrine-collections/en/2.3/index.html#selectable-methods:~:text=For%20collections%20that%20contain%20objects). 

The effects of this may be surprising. For example, with the ORM, it may be fine to filter entities based on the value of `private` or `protected` fields that have no getters. This works as long as a persistent collection is uninitialized. But as soon as it gets initialized, the `ArrayCollection` will require a getter method to be available.

Another (more rare) example is a getter method that does some type of type conversion, like having a `string` field with values like `'y'|'n'` internally but returning a `bool` value from the getter; or, more generally, every type mismatch between the return value of the getter and the field value. Yet another example may be getters that cause side effects 🙈.

#### Proposed solution

I discussed with @beberlei at the Doctrine hackathon that the primary use case for this library here was supporting the ORM/ODM use cases. This can be seen in places as `ClosureExpressionVisitor::getObjectFieldValue()` that take a `$field` parameter.

So, although this library here has nothing to do with ORM/ODM mapping, I want to add a migration path here that moves the `doctrine/collection` behaviour closer to the implementation realities of ORM/ODM. This means to ultimately use direct (reflection-based) field access only. This is also what @Ocramius already suggested in [this comment](https://github.com/doctrine/collections/pull/149#issuecomment-475982285).

For reference, here is a list of discussions around which style of accessors, getters, issers, public access etc. should be used or not used – in the future, the answer would be "only direct state (raw property value) matters".

* #276 
* #263 
* #149
* #135 
* #134 
* #95 
* #62 

#### Migration path

This feature is opt-in and will be activated by passing `accessRawFieldValues: true` to the `Criteria` constructor. The `Criteria` object is what is typically constructed by users in preparation for calling the `Selectable` API, so it seems to be a good fit.

By opting in through this flag, memory-based comparisons and sorting will use direct field access only. Not activating the feature triggers a deprecation notice. In the next major version, direct field access will be the only (default) behaviour.

The `$accessRawFieldValues` can be removed in the next major version (or, possibly, go through another round of deprecations in case when it is still passed, before being eventually removed).

#### Caveat

As explained in more detail in #487 (references here after for more better visibility), accessing fields through reflection may bypass initialization of proxy objects in the ORM/ODM. See #487 for more details.

#### Remaining edge case

Given an inheritance hierarchy of classes where a multiple classes feature a `private` field of the same name, the downmost field will be picked.

This may differ from Doctrine ORM behaviour when this field is not mapped at all in the ORM and another field (higher up the class hierarchy) is used as the mapped field instead.

We cannot solve this without having access to ORM/ODM mapping metadata at hand, which is not possible from within an `ArrayCollection` that is typically created as [a newable type](https://testing.googleblog.com/2008/10/to-new-or-not-to-new.html). We rather plan to discourage or even prevent this kind of setup (entity class hierarchy with different classes having fields of the same name) at some point when loading and validating metadata.

#### TODO

- [x] Docs update
- [x] PHP 8.4 raw property access (circumvent property hooks)
- [x] After merge and/or release, maybe revisit and update documentation remarks made in https://github.com/doctrine/orm/pull/11891/

### DIFF
--- a/src/Criteria.php
+++ b/src/Criteria.php
@@ -9,6 +9,7 @@ use Doctrine\Common\Collections\Expr\Expression;
 use Doctrine\Deprecations\Deprecation;
 
 use function array_map;
+use function func_get_arg;
 use function func_num_args;
 use function strtoupper;
 
@@ -38,9 +39,11 @@ class Criteria
      *
      * @return static
      */
-    public static function create()
+    public static function create(/* bool $accessRawFieldValues = false */): self
     {
-        return new static();
+        $accessRawFieldValues = 0 < func_num_args() ? func_get_arg(0) : false;
+
+        return new static(firstResult: 0, accessRawFieldValues: $accessRawFieldValues);
     }
 
     /**
@@ -67,8 +70,16 @@ class Criteria
         array|null $orderings = null,
         int|null $firstResult = null,
         int|null $maxResults = null,
+        private bool $accessRawFieldValues = false,
     ) {
-        $this->expression = $expression;
+        if (! $accessRawFieldValues) {
+            Deprecation::trigger(
+                'doctrine/collections',
+                'https://github.com/doctrine/collections/pull/472',
+                'Not enabling raw field value access for the Criteria matching API in %s is deprecated. Raw field access will be the only supported method in 3.0',
+                self::class,
+            );
+        }
 
         if ($firstResult === null && func_num_args() > 2) {
             Deprecation::trigger(
@@ -281,5 +292,11 @@ class Criteria
         $this->maxResults = $maxResults;
 
         return $this;
+    }
+
+    /** @internal */
+    public function isRawFieldValueAccessEnabled(): bool
+    {
+        return $this->accessRawFieldValues;
     }
 }
--- a/src/Expr/ClosureExpressionVisitor.php
+++ b/src/Expr/ClosureExpressionVisitor.php
@@ -6,11 +6,15 @@ namespace Doctrine\Common\Collections\Expr;
 
 use ArrayAccess;
 use Closure;
+use Doctrine\Deprecations\Deprecation;
+use ReflectionClass;
 use RuntimeException;
 
 use function array_all;
 use function array_any;
 use function explode;
+use function func_get_arg;
+use function func_num_args;
 use function in_array;
 use function is_array;
 use function is_scalar;
@@ -18,10 +22,13 @@ use function iterator_to_array;
 use function method_exists;
 use function preg_match;
 use function preg_replace_callback;
+use function sprintf;
 use function str_contains;
 use function str_ends_with;
 use function str_starts_with;
 use function strtoupper;
+
+use const PHP_VERSION_ID;
 
 /**
  * Walks an expression graph and turns it into a PHP closure.
@@ -31,6 +38,11 @@ use function strtoupper;
  */
 class ClosureExpressionVisitor extends ExpressionVisitor
 {
+    public function __construct(
+        private readonly bool $accessRawFieldValues = false,
+    ) {
+    }
+
     /**
      * Accesses the field of a given object. This field has to be public
      * directly or indirectly (through an accessor get*, is*, or a magic
@@ -40,18 +52,31 @@ class ClosureExpressionVisitor extends ExpressionVisitor
      *
      * @return mixed
      */
-    public static function getObjectFieldValue(object|array $object, string $field)
+    public static function getObjectFieldValue(object|array $object, string $field, /* bool $accessRawFieldValues = false */)
     {
+        $accessRawFieldValues = 3 <= func_num_args() ? func_get_arg(2) : false;
+
         if (str_contains($field, '.')) {
             [$field, $subField] = explode('.', $field, 2);
-            $object             = self::getObjectFieldValue($object, $field);
+            $object             = self::getObjectFieldValue($object, $field, $accessRawFieldValues);
 
-            return self::getObjectFieldValue($object, $subField);
+            return self::getObjectFieldValue($object, $subField, $accessRawFieldValues);
         }
 
         if (is_array($object)) {
             return $object[$field];
         }
+
+        if ($accessRawFieldValues) {
+            return self::getNearestFieldValue($object, $field);
+        }
+
+        Deprecation::trigger(
+            'doctrine/collections',
+            'https://github.com/doctrine/collections/pull/472',
+            'Not enabling raw field value access for %s is deprecated. Raw field access will be the only supported method in 3.0',
+            __METHOD__,
+        );
 
         $accessors = ['get', 'is', ''];
 
@@ -96,21 +121,52 @@ class ClosureExpressionVisitor extends ExpressionVisitor
         return $object->$field;
     }
 
+    private static function getNearestFieldValue(object $object, string $field): mixed
+    {
+        $reflectionClass = new ReflectionClass($object);
+
+        while ($reflectionClass && ! $reflectionClass->hasProperty($field)) {
+            $reflectionClass = $reflectionClass->getParentClass();
+        }
+
+        if ($reflectionClass === false) {
+            throw new RuntimeException(sprintf('Field "%s" does not exist in class "%s"', $field, $object::class));
+        }
+
+        $property = $reflectionClass->getProperty($field);
+
+        if (PHP_VERSION_ID >= 80400) {
+            return $property->getRawValue($object);
+        }
+
+        return $property->getValue($object);
+    }
+
     /**
      * Helper for sorting arrays of objects based on multiple fields + orientations.
      *
      * @return Closure
      */
-    public static function sortByField(string $name, int $orientation = 1, Closure|null $next = null)
+    public static function sortByField(string $name, int $orientation = 1, Closure|null $next = null, /* bool $accessRawFieldValues = false */)
     {
+        $accessRawFieldValues = 4 <= func_num_args() ? func_get_arg(3) : false;
+
+        if (! $accessRawFieldValues) {
+            Deprecation::trigger(
+                'doctrine/collections',
+                'https://github.com/doctrine/collections/pull/472',
+                'Not enabling raw field value access for %s is deprecated. Raw field access will be the only supported method in 3.0',
+                __METHOD__,
+            );
+        }
+
         if (! $next) {
             $next = static fn (): int => 0;
         }
 
-        return static function ($a, $b) use ($name, $next, $orientation): int {
-            $aValue = ClosureExpressionVisitor::getObjectFieldValue($a, $name);
-
-            $bValue = ClosureExpressionVisitor::getObjectFieldValue($b, $name);
+        return static function ($a, $b) use ($name, $next, $orientation, $accessRawFieldValues): int {
+            $aValue = ClosureExpressionVisitor::getObjectFieldValue($a, $name, $accessRawFieldValues);
+            $bValue = ClosureExpressionVisitor::getObjectFieldValue($b, $name, $accessRawFieldValues);
 
             if ($aValue === $bValue) {
                 return $next($a, $b);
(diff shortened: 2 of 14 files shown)


src/Criteria.php
```diff
@@ -33,14 +33,16 @@ class Criteria
...
     /**
      * Creates an instance of the class.
      *
      * @return static
      */
-    public static function create()
+    public static function create(bool $accessRawFieldValues = false): self
     {
-        return new static();
+        return new static(accessRawFieldValues: $accessRawFieldValues);
```
stof: this will break for child classes that override the constructor without defining a `accessRawFieldValues` constructor argument (which can totally happen if they override the constructor).

src/Expr/ClosureExpressionVisitor.php
```diff
@@ -96,21 +113,42 @@ public static function getObjectFieldValue(object|array $object, string $field)
...
+    private static function getNearestFieldValue(object $object, string $field): mixed
+    {
+        $reflectionClass = new \ReflectionClass($object);
+
+        while ($reflectionClass && !$reflectionClass->hasProperty($field)) {
+            $reflectionClass = $reflectionClass->getParentClass();
+        }
+
+        if (false === $reflectionClass) {
+            throw new RuntimeException(sprintf('Field "%s" does not exist in class %s', $field, get_class($object)));
```
stof:
```suggestion
            throw new RuntimeException(sprintf('Field "%s" does not exist in class "%s".', $field, get_class($object)));
```

src/Expr/ClosureExpressionVisitor.php
```diff
@@ -96,21 +115,42 @@ public static function getObjectFieldValue(object|array $object, string $field)
...
+
     /**
      * Helper for sorting arrays of objects based on multiple fields + orientations.
      *
      * @return Closure
      */
-    public static function sortByField(string $name, int $orientation = 1, Closure|null $next = null)
+    public static function sortByField(string $name, int $orientation = 1, Closure|null $next = null, bool $accessRawFieldValues = false)
     {
+        Deprecation::trigger(
```
stof: shouldn't this be triggered only based on  `false` value of the new argument ?

stof (commented):
I find it weird that so few tests are updated to use the non-deprecated usage, while also having no deprecations reported in the PHPUnit output. I think we need to improve the CI setup regarding deprecations.

The only tests that should trigger our own deprecations should be test covering legacy behaviors (basically tests that would be removed in the next major version when removing legacy code, without regressing code coverage)

src/Criteria.php
mpdude: Better now?

mpdude:
~This PR is now based on #473, so it includes deprecation checks in PHPUnit.~ That was merged, so I rebased.

Please advise about the coding standard issue.

greg0ire:
> Please advise about the coding standard issue.

I think in both cases, you are facing bugs, right? If yes, let's report them (if not reported by somebody else already) and ignore them with a comment.

mpdude:
https://github.com/PHPCSStandards/PHP_CodeSniffer/issues/734 - maybe no need to report anything?

mpdude:
@stof Could you give me another round of review here?

Thx!

src/Criteria.php
```diff
@@ -38,9 +38,12 @@ class Criteria
      *
      * @return static
      */
-    public static function create()
+    public static function create(bool $accessRawFieldValues = false): self
     {
-        return new static();
+        $new                       = new static();
+        $new->accessRawFieldValues = $accessRawFieldValues;
```
stof: wouldn't this delayed assignement trigger the deprecation in the constructor ?

tests/CriteriaTest.php
```diff
@@ -17,20 +17,29 @@ class CriteriaTest extends TestCase
...
+    public function testCreateLegacy(): void
     {
+        self::expectDeprecationWithIdentifier('https://github.com/doctrine/collections/pull/472');
+
         $criteria = Criteria::create();
 
         self::assertInstanceOf(Criteria::class, $criteria);
     }
 
+    public function testCreate(): void
```
stof, line 30: this test should explicitly verify that it has no deprecation with identifier `https://github.com/doctrine/collections/pull/472` (to potentially catch the mistake I reported).

the CI on this repo does not have a job failing on any deprecations.

greg0ire: :thinking: it should, since the hackathon: https://github.com/doctrine/collections/pull/473

mpdude: The `#[IgnoreDeprecations]` attribute around line 309 in `ArrayCollectionTestCase` causes that no deprecations are reported anymore for tests that are executed after it... Is that to be expected?

mpdude: I have no idea what's wrong here, but as to annotations, I get different results reported by PHPUnit depending on whether I just run `... phpunit` or e. g. `... phpunit tests/CriteriaTest.php`.

stof: this looks like a PHPUnit bug. this attribute should affect only that test, not subsequent ones.

mpdude: Hm, dunno... even without that:

```
$ php8.3 vendor/bin/phpunit tests/CriteriaTest.php
PHPUnit 12.4.1 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.26
Configuration: /var/www/doctrine-collections/phpunit.xml.dist

D................                                                 17 / 17 (100%)

Time: 00:00.013, Memory: 16.00 MB

1 test triggered 1 deprecation:

1) /var/www/doctrine-collections/vendor/doctrine/deprecations/src/Deprecation.php:208
Not enabling raw field value access for the Criteria matching API in Doctrine\Common\Collections\Criteria is deprecated. Raw field access will be the only supported method in 3.0 (Criteria.php:76 called by Criteria.php:43, https://github.com/doctrine/collections/pull/472, package doctrine/collections)

Triggered by:

* Doctrine\Tests\Common\Collections\CriteriaTest::testCreateLegacy
  /var/www/doctrine-collections/tests/CriteriaTest.php:20

OK, but there were issues!
Tests: 17, Assertions: 35, Deprecations: 1.
```

This test explicitly expects that deprecation. When I run all tests, nothing is printed.

mpdude: I am lost 😭

tests/CriteriaTest.php
```diff
@@ -17,20 +17,29 @@ class CriteriaTest extends TestCase
 {
     use VerifyDeprecations;
 
-    public function testCreate(): void
+    public function testCreateLegacy(): void
```
greg0ire, line 21: I think you will need to use `#[IgnoreDeprecations]` on every test that calls `expectDeprecationWithIdentifier`

src/Criteria.php
mpdude: Yes, it should have – glad you spotted it. Now fixed, and would have been reported after the changes from #479.

tests/CriteriaTest.php
mpdude: The problem was that the Deprecations library was trying to de-duplicate deprecation notices by default. That meant that depending on which set of tests are being executed, particular tests would be the first to raise the notice or not, leading to surprising results.

#479 turns of deduplication and fixes the notices popping up for the first time.

tests/bootstrap.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+declare(strict_types=1);
+
+use Doctrine\Deprecations\Deprecation;
+
+Deprecation::withoutDeduplication();
```
greg0ire, line 7: FYI, I'm working on making this easier: https://github.com/doctrine/deprecations/pull/94

tests/CriteriaTest.php
stof: We should probably implement a way to automatically default to no deduplication when running tests, to avoid such issues.

mpdude:
Feedback and deprecations addressed

greg0ire:
Since this introduces deprecations, you should target 2.4.x

src/Expr/ClosureExpressionVisitor.php
```diff
@@ -31,6 +36,11 @@
  */
 class ClosureExpressionVisitor extends ExpressionVisitor
 {
+    public function __construct(
+        private readonly bool $accessRawFieldValues = false,
+    ) {
```
greg0ire, line 43: Should there be a deprecation here as well?

src/Expr/ClosureExpressionVisitor.php
```diff
@@ -40,19 +50,30 @@ class ClosureExpressionVisitor extends ExpressionVisitor
      *
      * @return mixed
      */
-    public static function getObjectFieldValue(object|array $object, string $field)
+    public static function getObjectFieldValue(object|array $object, string $field, bool $accessRawFieldValues = false)
```
greg0ire: This class is not final nor internal, so this is a breaking change. You could use `func_get_args()` to address this issue.

Also, why not use `$this->accessRawFieldValues` throughout this method instead?

Same goes for the public method below.

src/Criteria.php
```diff
@@ -38,9 +38,9 @@ class Criteria
      *
      * @return static
      */
-    public static function create()
+    public static function create(bool $accessRawFieldValues = false): self
```
greg0ire: This is a breaking change, you should use `func_get_args()` IMO

https://3v4l.org/Z4ZnM

src/Expr/ClosureExpressionVisitor.php
mpdude: I think we don't need it here: This member in ClosureExpressionVisitor is only relevant for the `walkComparison()` method, where the flag will be passed on to the static methods. It will trigger notices there.

src/Expr/ClosureExpressionVisitor.php
mpdude: As to 1.), I'll look into it. 
As to 2.), it's a static method so we cannot access `$this`

mpdude:
Rebased on top of #480 to make it target 2.4.x. Change will look smaller once #480 is merged.

src/Criteria.php
mpdude: Well spotted! fixed

beberlei (dismissed):

beberlei:
Great work cleaning this up

mpdude:
Now rebased against the up-to-date 2.4.x branch

greg0ire (approved):

greg0ire:
Thanks @mpdude !

src/Criteria.php
```diff
@@ -67,8 +70,16 @@ public function __construct(
...
+            Deprecation::trigger(
+                'doctrine/collections',
+                'https://github.com/doctrine/collections/pull/472',
+                'Not enabling raw field value access for the Criteria matching API in %s is deprecated. Raw field access will be the only supported method in 3.0',
+                self::class,
+            );
+        }
 
         if ($firstResult === null && func_num_args() > 2) {
             Deprecation::trigger(
```
W0rma, line 85: I just noticed that this deprecation will always be triggered if f.e. `new Criteria(accessRawFieldValues: true)` is used.

That's because of the nature of `func_num_args()` as explained in https://www.php.net/manual/en/function.func-num-args.php:

> As of PHP 8.0.0, the func_*() family of functions is intended to be mostly transparent with regard to named arguments, by treating the arguments as if they were all passed positionally, and missing arguments are replaced with their defaults. This function ignores the collection of unknown named variadic arguments. Unknown named arguments which are collected can only be accessed through the variadic parameter.

See also https://3v4l.org/5X4Xu#v8.4.14

Users can work around that by either using `new Criteria(firstResult: 0, accessRawFieldValues: true)` or  `Criteria::create(true)`.

Just FYI if other users might get confused about that as well.

PR #473, opened 2025-10-08
Configure PHPUnit to fail on doctrine deprecations

greg0ire:

PR #479, opened 2025-10-22
Fix reporting of deprecations in tests

mpdude:
Because the de-duplication of deprecation notices was enabled by default, not all actual deprecations were reported after #473.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -4,6 +4,7 @@
      colors="true"
      beStrictAboutOutputDuringTests="true"
      failOnDeprecation="true"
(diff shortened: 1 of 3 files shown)


mpdude:
@greg0ire May I ask for your review? I cannot use the GH review request feature here.

greg0ire (approved):

mpdude:
It would be nice if we could merge this as-is, so that #472 can proceed.

Once https://github.com/doctrine/deprecations/pull/94 is merged and released, we can still improve here.

SenseException (approved):

greg0ire:
Thanks @mpdude !

PR #486, opened 2025-10-26
Remove the `accessRawFieldValues` parameter which is a no-op as of 3.0

mpdude:
This removes the `accessRawFieldValues` parameter that was added in #472 in the 3.0.0 branch. 

Since this major version, fields values are always as raw values by means of reflection, so the parameter was a no-op.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,6 +8,17 @@ awareness about deprecated code.
 
 # Upgrade to 3.0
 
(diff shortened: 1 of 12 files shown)


greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @mpdude !

PR #488, opened 2025-10-28
Add a note that #472 may leave proxies uninitialized

### LINKED ISSUE #487: Filtering API: Accessing fields through reflection does not initialize proxies
When enabling the reflection-based access of fields (#472), proxy objects might not get initialized. This can lead to subtle changes in behavior.

Before #472 (or without activating this feature), access through getters or public properties triggered typical proxy initialization logic.

#### Motivating example

I am not sure whether a `Collection` as  provided by Doctrine ORM for to-many-associations can contain uninitalized proxies in the first place. In my particular case, the proxies come from a situation as follows:

I have an Entity `Publisher`, with a to-many association on `Book`, with a to-many association on `Author`.

There is a method `Publisher::getAuthorsOrderedByYearOfBirth()` along the lines of:

```php
return $this->books
    ->map(static fn (Book $b): Author => $b->getAuthor())
    ->matching(Criteria::create(true)->orderBy(['birthdate' => Order::Ascending]));
```

(Ignore potentially duplicate authors for the sake of simplicity.)

Here, `Book::getAuthor()` may be the source of an uninitialized `Author` proxy (created by the ORM). Accessing the `birthdate` field through reflection on these proxies leads to `null` values without triggering initialization, effectively returning the result with not particular sort order applied.

#### Possible solution (1)

We could include proxy detection and initialization logic in `ClosureExpressionVisitor::getObjectFieldValue()`, which might be as simple as

```php
if ($object instanceof \Doctrine\Persistence\Proxy) {
    $object->__load();
}
```

I assume this would fix the problem for all ORM/ODM scenarios pre native lazy objects. Rumor has it that native lazy objects will be initialized through reflection-based access as well, so no problem there.

Pro:
* Might transparently fix the problem for most of the practical situations.

Downsides:
* Very much tailored to the ORM/ODM use cases. 
* `doctrine/persistence` is currently not a dependency of this library here. Unsure whether it needs to be or should become for that check. On the short run technically no, we don't need it for the `instanceof` check. Conceptually, maybe yes – theoretically, the next major version of it might not include the `__load` method anymore.

#### Possible solution (2)

Raise awareness and tell users to fix the problem in userland:

```php
return $this->books
    ->map(static function (Book $b): Author {
        $author = $b->getAuthor();
        if ($author instanceof Proxy) {
            $author->__load();
        }
    )
    ->matching(Criteria::create(true)->orderBy(['birthdate' => Order::Ascending]);
```

Pro:
* Nothing to do here (apart from docs update, maybe)

Downside:
* Higher WTF ratios, bad DX, subtle/surprising bugs
* Slight increase in global code pollution through boilerplate code added.

mpdude:
This adds a note to the UPGRADE file that opting in to #472 may cause a change in behaviour for proxy objects (#487).

Closes #487.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -12,11 +12,16 @@ awareness about deprecated code.
 
 Starting with the next major version, the only way to access data when using the criteria filtering 
 API is through direct (reflection-based) access at properties directly, also bypassing property hooks.
-This is to ensure consistency with how the ORM/ODM work. See https://github.com/doctrine/collections/pull/472 for
+This is to ensure consistency with how the ORM/ODM works. See https://github.com/doctrine/collections/pull/472 for
 the full motivation.
 
 To opt-in to the new behaviour, pass `true` for the `$accessRawFieldValues` parameter when creating a `Criteria`
-object through either `Doctrine\Common\Collections\Criteria::create()` or when calling the `Doctrine\Common\Collections\Criteria` constructor. 
+object through either `Doctrine\Common\Collections\Criteria::create()` or when calling the `Doctrine\Common\Collections\Criteria` constructor.
+
+Be aware that switching to reflection-based field access may prevent ORM or ODM proxy objects
+becoming initialized, since their triggers (like calling public methods) are bypassed. That might lead
+to `null` values being read from such objects, which may cause wrong filtering or sorting results.
+See https://github.com/doctrine/collections/issues/487 for more details on when this may happen.
 
 # Upgrade to 2.2
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -19,7 +19,7 @@ To opt-in to the new behaviour, pass `true` for the `$accessRawFieldValues` para
 object through either `Doctrine\Common\Collections\Criteria::create()` or when calling the `Doctrine\Common\Collections\Criteria` constructor.
 
 Be aware that switching to reflection-based field access may prevent ORM or ODM proxy objects
-becoming initialized, since their triggers (like calling public methods) are bypassed. That might lead
+becoming initialized, since their triggers (like calling public methods) are bypassed. This might lead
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -12,11 +12,16 @@ awareness about deprecated code.
...
-This is to ensure consistency with how the ORM/ODM work. See https://github.com/doctrine/collections/pull/472 for
+This is to ensure consistency with how the ORM/ODM works. See https://github.com/doctrine/collections/pull/472 for
 the full motivation.
 
 To opt-in to the new behaviour, pass `true` for the `$accessRawFieldValues` parameter when creating a `Criteria`
-object through either `Doctrine\Common\Collections\Criteria::create()` or when calling the `Doctrine\Common\Collections\Criteria` constructor. 
+object through either `Doctrine\Common\Collections\Criteria::create()` or when calling the `Doctrine\Common\Collections\Criteria` constructor.
+
+Be aware that switching to reflection-based field access may prevent ORM or ODM proxy objects
+becoming initialized, since their triggers (like calling public methods) are bypassed. That might lead
```
alcaeus:
```suggestion
becoming initialized, since their triggers (like calling public methods) are bypassed. This might lead
```

UPGRADE.md
```diff
@@ -12,11 +12,16 @@ awareness about deprecated code.
...
+to `null` values being read from such objects, which may cause wrong filtering or sorting results.
```
alcaeus, line 23: We should suggest an alternative since lazy objects don't suffer from this issue:

```suggestion
to `null` values being read from such objects, which may cause wrong filtering or sorting results.
To avoid this issue, use native lazy objects added in PHP 8.4.
```

alcaeus (commented):
LGTM with some wording changes

alcaeus (approved):

greg0ire:
Thanks @mpdude !

Issue #155, opened 2018-06-28
ArrayCollection should have initilized $elements property


Arkemlar:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | v1.5.0 (doctrine/collections)

#### Summary

ArrayCollection class has unitialised private `$elements` property which leads to `null` value in it in some cases. 
It needs to be initialized:
`private $elements;` => `private $elements = [];`

#### Current behavior

In case of deserialization the value of the property is not safty - it might be `null`! As the result - I getting error exception when calling ArrayCollection's methods which contains inner calls to php functions with array only arguments (e.g. `getValues()` calls ` array_values()`).

#### How to reproduce

1. Extend ArrayCollection, and implement /Serializable interface (add custom seialize and userialize methods).
```
class ExtendedArrayCollectionWithCustomSerializer extends ArrayCollection implements Serializable
{
    public function serialize()
    {
        return json_encode($this->getKeys());
    }

    public function unserialize($serialized)
    {
        // $this->clear(); // Workaround to force $elements initialisation

        foreach (json_decode($serialized) as $value) {
            parent::add($value);
        }
    }
}
```

2. Create an **empty** ArrayCollection instance, then serialize and userialize it, then check unserialized instance - it will have `null` value in `$elements` property. 
3. Call something like `getValues()` you will get an error: `Warning: array_values() expects parameter 1 to be array, null given`

```
$collection = new ExtendedArrayCollectionWithCustomSerializer();
$collection = unserialize(serialize($collection));
// At this point $collection->elements === null
$collection->getValues();  // ErrorException there: Warning: array_values() expects parameter 1 to be array, null given
```

#### Expected behavior

Have `$elements` property either protected or initilized so we can write safety deserialization code (and therefore never get `null` as `$elements` property value)

Ocramius:
@Arkemlar can you maybe make an example to be added to https://github.com/doctrine/collections/blob/ab12b39de988ee3d80bf2d5f16b1161ee5fb2529/tests/Doctrine/Tests/Common/Collections/ArrayCollectionTest.php?

Arkemlar:
Yep, next week.

PR #503, opened 2026-01-05
Mark all classes as final

greg0ire:
They will be hard-final in 3.0

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,6 +6,16 @@ awareness about deprecated code.
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
 
+# Upgrade to 2.5
+
+Extending the following classes is deprecated and will no longer be possible in 3.0:
+
+- `Doctrine\Common\Collections\Criteria`
+- `Doctrine\Common\Collections\Expr\ClosureExpressionVisitor`
+- `Doctrine\Common\Collections\Expr\Comparison`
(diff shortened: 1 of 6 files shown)


src/ArrayCollection.php
```diff
@@ -48,6 +48,7 @@
  * @template-implements Collection<TKey,T>
  * @template-implements Selectable<TKey,T>
  * @phpstan-consistent-constructor
+ * @final
```
greg0ire: This class is extended in tests, by `SerializableArrayCollection`, since #155

greg0ire: According to https://github.com/doctrine/collections/issues/155, extending `ArrayCollection` is something people seem to do, so let's skip this one I'd say.

derrabus (dismissed):

greg0ire:
:thinking: technically, this is a deprecation. Let me document this in the upgrade guide then.

derrabus:
I don't know if we've discussed this before, but Symfony uses the syntax `@final since X.Y` in PHPDoc blocks to indicate that the class should be considered final by downstream code, but we cannot apply changes to it that would be breaking ones for a non-final class (e.g. add parameters to methods) until version (X+1).0 is released. And I kinda like that strategy.

greg0ire:
We've attempted to discuss this during the Hackathon. I like it as well, but I'm not sure everyone does. Let's discuss this on Slack?

greg0ire:
Clarified internally, let's use it.

SenseException (approved):

PR #510, opened 2026-01-07
Use native PHP keyword to make classes final

greg0ire:


SenseException (approved):
I wonder how often `Criteria` and `ExpressionBuilder` were extended in userland.

derrabus (approved):

greg0ire:
If it becomes an issue we can partially revert.

PR #511, opened 2026-01-08
Add missing @final annotation

greg0ire:
This was overlooked in 3c464a2


greg0ire:
Technically, should go in the next minor release but… can we make an exception here :pleading_face: ?

SenseException (approved):
`@final` was compared to a deprecation, but I'd like to hear a second opinion.

GromNaN (approved):

PR #513, opened 2026-01-08
Add Override attribute where applicable

greg0ire:
It makes it more obvious to humans or tools what is overriding/implementing something and what is not.

### DIFF
--- a/src/AbstractLazyCollection.php
+++ b/src/AbstractLazyCollection.php
@@ -6,6 +6,7 @@ namespace Doctrine\Common\Collections;
 
 use Closure;
 use LogicException;
+use Override;
 use Traversable;
 
 /**
@@ -27,6 +28,7 @@ abstract class AbstractLazyCollection implements Collection
 
     protected bool $initialized = false;
 
+    #[Override]
     public function count(): int
     {
         $this->initialize();
@@ -34,6 +36,7 @@ abstract class AbstractLazyCollection implements Collection
         return $this->collection->count();
     }
 
+    #[Override]
     public function add(mixed $element): void
     {
         $this->initialize();
@@ -41,12 +44,14 @@ abstract class AbstractLazyCollection implements Collection
         $this->collection->add($element);
     }
 
+    #[Override]
     public function clear(): void
     {
         $this->initialize();
         $this->collection->clear();
     }
 
+    #[Override]
     public function contains(mixed $element): bool
     {
         $this->initialize();
@@ -54,6 +59,7 @@ abstract class AbstractLazyCollection implements Collection
         return $this->collection->contains($element);
     }
 
+    #[Override]
     public function isEmpty(): bool
     {
         $this->initialize();
@@ -61,6 +67,7 @@ abstract class AbstractLazyCollection implements Collection
         return $this->collection->isEmpty();
     }
 
+    #[Override]
     public function remove(string|int $key): mixed
     {
         $this->initialize();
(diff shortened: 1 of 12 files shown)


derrabus:
Is there any tooling that complains if we forget to set that attribute anywhere? Otherwise, maintaining consistency will be a nightmare. 🙈

greg0ire:
@derrabus good point, and yes! https://phpstan.org/config-reference#checkmissingoverridemethodattribute

Let me try that.

greg0ire:
It actually caught a few I forgot (shouldn't have trusted an AI with this I guess). The LLM assumed interface would not override method, but we do because we need to add static analysis annotations.

```
 ------ -------------------------------------------------------------------------------------------------------------------
  Line   src/Collection.php
 ------ -------------------------------------------------------------------------------------------------------------------
  88     Method Doctrine\Common\Collections\Collection::map() overrides method
         Doctrine\Common\Collections\ReadableCollection<TKey of (int|string),T>::map() but is missing the #[\Override] att
         ribute.
         🪪  method.missingOverride
  98     Method Doctrine\Common\Collections\Collection::filter() overrides method
         Doctrine\Common\Collections\ReadableCollection<TKey of (int|string),T>::filter() but is missing the #[\Override]
         attribute.
         🪪  method.missingOverride
  110    Method Doctrine\Common\Collections\Collection::partition() overrides method
         Doctrine\Common\Collections\ReadableCollection<TKey of (int|string),T>::partition() but is missing the #[\Overrid
         e] attribute.
         🪪  method.missingOverride
 ------ -------------------------------------------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------------------------------------------
  Line   tests/StaticAnalysis/CustomCollection.php
 ------ -------------------------------------------------------------------------------------------------------------------
  32     Method Doctrine\Tests\Common\Collections\StaticAnalysis\CustomCollection::filter() overrides method
         Doctrine\Common\Collections\Collection<TKey of (int|string),T of object>::filter() but is missing the #[\Override
         ] attribute.
         🪪  method.missingOverride
  42     Method Doctrine\Tests\Common\Collections\StaticAnalysis\CustomCollection::partition() overrides method
         Doctrine\Common\Collections\Collection<TKey of (int|string),T of object>::partition() but is missing the #[\Overr
         ide] attribute.
         🪪  method.missingOverride
 ------ -------------------------------------------------------------------------------------------------------------------


 [ERROR] Found 5 errors
```

greg0ire:
Requesting reviews from lots of different people before I do the same on every library.

src/AbstractLazyCollection.php
```diff
@@ -92,6 +102,7 @@ public function get(string|int $key): mixed
     /**
      * {@inheritDoc}
      */
```
GromNaN: The `{@inheritDoc}` alone seems redundant with the new attribute. It should be removed as the tools already read the doc from the parent method.

GromNaN (commented):
Even though it is very verbose, the advantage I see in this attribute is that if you delete a method in a parent class, you get an error to delete it in the child classes as well.

src/AbstractLazyCollection.php
greg0ire: That's a great point !

morozov (approved):

GromNaN (approved):

derrabus (approved):
Not exactly a fan of this attribute yet, but let's see where this takes us.

PR #515, opened 2026-01-12
Avoid triggering deprecation on named argument

### LINKED ISSUE #489: Incorrect deprecation notice
### Bug Report

|    Q                                        |   A
|-------------------------------------------- | ------
| Version                                     | 2.4.0
| Previous Version if the bug is a regression | 2.3.0 (probably)

#### Summary

Valid code is triggering a deprecation notice.

#### Current behavior

This code is triggering this notice.

`Passing null as $firstResult to the constructor of Doctrine\Common\Collections\Criteria is deprecated. Pass 0 instead or omit the argument. (Criteria.php:85 called by Seat.php:530, https://github.com/doctrine/collections/pull/311, package doctrine/collections)`

with this code:

```php
new Criteria(orderings: ['startDate' => Order::Descending], accessRawFieldValues: true)
```

I think the counting of the function args isn't working as expected

```php
        if ($firstResult === null && func_num_args() > 2) {
```

#### Expected behavior

This code should not be triggering this notice.

greg0ire:
Our detection of whether $firstResult based on func_num_args() was unreliable with named arguments. The following prints `int(2)`:

```php
<?php
(function($a=42, $b=42) {
    var_dump(func_num_args());
})(b: 42);
```

Instead, let's change the default value for `$firstResult` to something very unlikely, and assume nobody will be passing that.

Fixes #489

### DIFF
--- a/src/Criteria.php
+++ b/src/Criteria.php
@@ -63,12 +63,13 @@ class Criteria
     /**
      * Construct a new Criteria.
      *
+     * @param int|null                         $firstResult
      * @param array<string, string|Order>|null $orderings
      */
     public function __construct(
         private Expression|null $expression = null,
         array|null $orderings = null,
-        int|null $firstResult = null,
+        int|Placeholder|null $firstResult = Placeholder::NotSpecified,
         int|null $maxResults = null,
         private bool $accessRawFieldValues = false,
     ) {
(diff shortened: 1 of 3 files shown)


GromNaN (approved):
Excellent use of single-value enum.

Crell:
Muhahaha, it's spreading! :smile:

GromNaN:
We have the same in the MongoDB library: https://github.com/mongodb/mongo-php-library/blob/v2.x/src/Builder/Type/Optional.php

PR #518, opened 2026-01-14
Implement Selectable

greg0ire:
This unlocks having `ReadableCollection` extend `Selectable` in the next
major version.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,6 +6,16 @@ awareness about deprecated code.
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
 
+# Upgrade to 2.6
+
+When extending `Doctrine\Common\Collections\AbstractLazyCollection`, the
+backing collection initialized in `doInitialize()` must implement
+`Doctrine\Common\Collections\Selectable`. Initializing with a collection that
+does not implement `Selectable` is deprecated and will throw an exception in 3.0.
+
+Also, implementing `ReadableCollection` without implementing `Selectable`
+deprecated and will be an error in 3.0.
+
 # Upgrade to 2.5
 
 Extending the following classes is deprecated and will no longer be possible in 3.0:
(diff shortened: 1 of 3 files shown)


UPGRADE.md
```diff
@@ -6,6 +6,16 @@ awareness about deprecated code.
...
 
+# Upgrade to 2.6
+
+When extending `Doctrine\Common\Collections\AbstractLazyCollection`, the
+backing collection initialized in `doInitialize()` must implement
+`Doctrine\Common\Collections\Selectable`. Initializing with a collection that
+does not implement `Selectable` is deprecated and will throw an exception in 3.0.
+
+Also, implementing `ReadableCollection` without implementing `Selectable`
+deprecated and will be an error in 3.0.
```
greg0ire, line 17: I could not find a way to enforce this particular deprecation. In particular, I did not find an annotation that would cause either PHPStan or Symfony's `DebugErrorHandler` to complain about this.

src/AbstractLazyCollection.php
```diff
@@ -411,4 +425,18 @@ protected function initialize()
      * @return void
      */
     abstract protected function doInitialize();
+
+    /**
+     * {@inheritDoc}
+     */
+    public function matching(Criteria $criteria): ReadableCollection&Selectable
```
stof: this should not add a native return type in 2.x to avoid introducing a stricter parent type than `Selectable` for child classes (for instance the PersistentCollection of the ORM, which currently has `Collection` as return type, which is not a subtype of `Selectable` for now).
The native return type should be added only in 3.0 (at the same time than adding it in the Selectable interface).

UPGRADE.md
GromNaN: You can change the property type in the phpdoc to `@phpstan-var Collection<TKey,T>&Selectable|null`

greg0ire: Which property are you referring to? Here, what I would like to happen is that any collection implementing `ReadableCollection` but not implementing `Selectable` causes a static analysis error or a deprecation. You don't need to have a property to implement `ReadableCollection`.

stof (approved):
The ODM will need to implement Selectable in its persistent collections (/cc @GromNaN and @alcaeus).

PR #522, opened 2026-01-15
Make ReadableCollection extend Selectable

greg0ire:


src/AbstractLazyCollection.php
```diff
@@ -23,10 +22,9 @@ abstract class AbstractLazyCollection implements Collection, Selectable
     /**
      * The backed collection to use
      *
-     * @phpstan-var Collection<TKey,T>|null
-     * @var Collection<mixed>|null
+     * @var (Collection<TKey,T>&Selectable<TKey,T>)|null
      */
-    protected Collection|null $collection;
+    protected (Collection&Selectable)|null $collection;
```
stof: this is equivalent to just `Collection` now, as `Collection` extends `ReadableCollection` which now extends `Selectable`

derrabus (approved):

stof (approved):

PR #542, opened 2026-04-22
Deprecate Order enum

greg0ire:
PHP 8.6 provides a native `\SortDirection` enum for this exact purpose. See https://wiki.php.net/rfc/sort_direction_enum

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,10 +1,25 @@
 Note about upgrading: Doctrine uses static and runtime mechanisms to raise
 awareness about deprecated code.
 
-- Use of `@deprecated` docblock that is detected by IDEs (like PHPStorm) or
+- Use of `#[Deprecated]` attribute that is detected by IDEs (like PHPStorm) or
   Static Analysis tools (like Psalm, phpstan)
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
+
+# Upgrade to 3.1
+
+## Deprecated `Order` enum
+
+PHP 8.6 provides a native `\SortDirection` enum that should be used instead of
+the `Doctrine\Common\Collections\Order` enum. The `Order` enum is deprecated
+and will be removed in 4.0.
+
+`\SortDirection` is polyfilled by the `symfony/polyfill-php86` package, that we
+require.
+
+As a consequence, `Criteria::orderings()`, which returns an array of `Order`
+instances, is deprecated in favor of `Criteria::getOrderings()`, which returns
+an array of `\SortDirection` instances.
 
 # Upgrade to 3.0
 
--- a/src/Criteria.php
+++ b/src/Criteria.php
@@ -4,9 +4,11 @@ declare(strict_types=1);
 
 namespace Doctrine\Common\Collections;
 
+use Deprecated;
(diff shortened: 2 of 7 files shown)


src/Criteria.php
```diff
@@ -144,25 +146,50 @@ public function getWhereExpression(): Expression|null
         return $this->expression;
     }
 
+    /**
+     * Gets the current orderings of this Criteria.
+     *
+     * @return array<string, SortDirection>
+     */
+    public function getOrderings(): array
```
greg0ire, line 154: Note that I'm reusing the name of a method that existed in the past here, and got deprecated in favor of `orderings` for the same reason I'm doing this here.

bobvandevijver:
I love this new development, and I can only hope that this will also find it's way to other Doctrine packages!

src/Order.php
```diff
@@ -4,6 +4,7 @@
 
 namespace Doctrine\Common\Collections;
 
+/** @deprecated use \SortDirection instead */
```
greg0ire, line 7: `#[Deprecated]` does not appear to be supported on enums yet: https://3v4l.org/ON3VU#v8.4.14

src/Criteria.php
greg0ire: Maybe I should pick another name, so that it's possible to port this to 2.8 as a forward-compatibility improvement 🤔 

It would avoid having people on 2.x move to `Order`, only to find they have to move to `SortDirection` once they upgrade to 3.x

greg0ire: If I use `getOrderedBy()` as the new name, I think it would also make sense to rename `Criteria::orderings` to be consistent, but then doing the same for arguments of e.g. `Criteria::__construct()` would be a breaking change because of named arguments, so we would end up with a naming inconsistency.

UPGRADE.md
```diff
@@ -1,11 +1,26 @@
...
+# Upgrade to 3.1
+
+## Deprecated `Order` enum
+
+PHP 8.6 provides a native `\SortDirection` enum that should be used instead of
+the `Doctrine\Common\Collections\Order` enum. The `Order` enum is deprecated
+and will be removed in 4.0.
+
+`\SortDirection` is polyfilled by the `symfony/polyfill-php86` package, that we
+require.
```
SenseException, line 18: I don't find `symfony/polyfill-php86` in the target branch in composer.

bobvandevijver: That's why this PR adds it

UPGRADE.md
```diff
@@ -1,11 +1,26 @@
...
+
+As a consequence, `Criteria::orderings()`, which returns an array of `Order`
+instances, is deprecated in favor of `Criteria::getOrderings()`, which retursn
```
Hanmac: Small typo
```suggestion
instances, is deprecated in favor of `Criteria::getOrderings()`, which returns
```

UPGRADE.md
SenseException: :facepalm: I oversaw that one

SenseException (approved):

src/Criteria.php
greg0ire: Let's stay away from the naming inconsistency. Sorry for the users :bow: The double migration won't be too painful I hope.

PR #552, opened 2026-09-05
fix: use value equality for sortByField tie detection

dualfroz:
Fixes #437.

## Problem

`Criteria::orderBy()` with multiple fields silently drops every field after the
first whenever the first field's value is an object (e.g. `DateTimeImmutable`,
as in the issue's repro). Two records with the same `validFrom` but different
`id` are not ordered by `id` at all — and the resulting order depends on the
input order of the array being sorted, which a comparator should never do.

`mpdude` confirmed the same root cause on the issue and linked the duplicate
#136 and an earlier unmerged attempt (#454, which added an opt-in
`treatDateTimeAsScalar()` flag scoped to `DateTimeInterface` rather than fixing
the tie-detection itself).

## Root cause

In `sortByField()` the tie check uses `===` (identity), while the ordering
decision one line below uses `>` (value comparison):

```php
if ($aValue === $bValue) {   // identity
    return $next($a, $b);
}

return ($aValue > $bValue ? 1 : -1) * $orientation;   // value comparison
```

Two distinct `DateTimeImmutable` instances for the same instant are `==` but
never `===`, so the tie branch is never taken: `$next()` — the next `orderBy()`
field — is never consulted, and the expression below degrades to an arbitrary
`1`/`-1` for values that are in fact equal, which is not antisymmetric.

## Fix

Detect the tie with the same comparison that already decides the ordering:

```php
$comparison = $aValue <=> $bValue;

if ($comparison === 0) {
    return $next($a, $b);
}

return $comparison * $orientation;
```

`<=>` is built from the same `<` / `==` / `>` semantics the previous code
already relied on, so this is behaviour-preserving for every type it handled
before, and the comparator becomes antisymmetric by construction. Scope is
`sortByField()` plus the tests below; no public API change.

## Tests

Two cases added to `ClosureExpressionVisitorTest`: one asserting the secondary
field is consulted when the primary values are equal-but-not-identical, one
sorting the same pair from both possible input orderings and asserting the
result is identical. Both fail before the change and pass after it.

Reproduced through the public API as in the issue — before the fix the two
input orderings disagree (`6` vs `5`), after it they agree (`6` and `6`).

---

Targeted at `2.6.x` because `3.1.x` requires PHP ^8.4; the same code and bug are
present on both. Happy to retarget if you'd prefer it elsewhere.

### DIFF
--- a/src/Expr/ClosureExpressionVisitor.php
+++ b/src/Expr/ClosureExpressionVisitor.php
@@ -170,11 +170,13 @@ class ClosureExpressionVisitor extends ExpressionVisitor
             $aValue = ClosureExpressionVisitor::getObjectFieldValue($a, $name, $accessRawFieldValues);
             $bValue = ClosureExpressionVisitor::getObjectFieldValue($b, $name, $accessRawFieldValues);
 
-            if ($aValue === $bValue) {
+            $comparison = $aValue <=> $bValue;
+
+            if ($comparison === 0) {
                 return $next($a, $b);
             }
 
-            return ($aValue > $bValue ? 1 : -1) * $orientation;
+            return $comparison * $orientation;
         };
     }
 
--- a/tests/ClosureExpressionVisitorTest.php
+++ b/tests/ClosureExpressionVisitorTest.php
@@ -6,6 +6,7 @@ namespace Doctrine\Tests\Common\Collections;
 
 use ArrayAccess;
 use ArrayIterator;
+use DateTimeImmutable;
 use Doctrine\Common\Collections\Expr\ClosureExpressionVisitor;
 use Doctrine\Common\Collections\Expr\Comparison;
 use Doctrine\Common\Collections\Expr\CompositeExpression;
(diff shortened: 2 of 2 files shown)


greg0ire:
> phpstan reports the same 5
pre-existing findings on both the unmodified branch and this branch (an
8.4-only ReflectionProperty::getRawValue() call flagged under the local
PHP 8.3 toolchain, plus 3 findings in unrelated, untouched test-fixture
classes) -- none introduced by this change.

You should run PHPStan the way it is run in the CI.


> No PR was opened against doctrine/collections.

What am I looking at then?

src/Expr/ClosureExpressionVisitor.php
```diff
@@ -170,11 +170,16 @@ public static function sortByField(string $name, int $orientation = 1, Closure|n
             $aValue = ClosureExpressionVisitor::getObjectFieldValue($a, $name, $accessRawFieldValues);
             $bValue = ClosureExpressionVisitor::getObjectFieldValue($b, $name, $accessRawFieldValues);
 
-            if ($aValue === $bValue) {
+            // Detect ties with the same comparison that orders the values below: a
+            // strict "===" check misses value-equal objects such as two
```
greg0ire: Please refrain from contrasting new code with code it replaces. One shouldn't need to look at the VCS history to understand what you are referring to.

greg0ire:
>  Commit: 990f3f475a8b40d51275352598f963fb436c6175

What is this about?

greg0ire:
In https://github.com/doctrine/collections/pull/454, @mpdude was cautious about BC: 

> Of course, doing so would be a BC break with regard to filtering and sorting. Thus, this PR adds a new Criteria::treatDateTimeAsScalar() method that flips Criteria instances into the new mode of operation. That flag is passed on later to ClosureExpressionVisitor where the actual comparisons take place.

Should something similar be done here?

dualfroz:
Sorry for the noise, the description had leftover notes from my local workflow that shouldn't have been in the PR body. I've cleaned it up, should be fine now. My bad, sorry again.

dualfroz:
Thanks — three answers.

**PHPStan.** You were right that I ran it wrong. Run with the repository's own config (`level: 8`, `paths: src, tests/StaticAnalysis`), there is exactly one error:

```
 ------ ----------------------------------------------------------------
  Line   src/Expr/ClosureExpressionVisitor.php
 ------ ----------------------------------------------------------------
  141    Call to an undefined method ReflectionProperty::getRawValue().
 ------ ----------------------------------------------------------------
```

It reproduces identically on `2.6.x` at `1e91b75` without my changes. Line 141 is the existing `PHP_VERSION_ID >= 80400` branch, and `getRawValue()` only exists on 8.4, so it is an artefact of analysing on an older PHP locally. CI analyses on 8.4, where it does not fire.

**Backward compatibility.** #454 changed how `Comparison::EQ`/`NEQ` are evaluated, so it altered `filter()` as well as sorting — a user filtering by a `DateTimeInterface` would start matching where they previously did not. That is a real semantic change to well-defined behaviour, so the opt-in flag made sense.

This change does not touch `walkComparison()`; `Comparison::EQ` still uses `===`, so filtering is untouched. Only the tie check inside `sortByField()` changes, and only for pairs where the previous comparator returned `-1` in *both* directions:

| values | before `cmp(a,b)` / `cmp(b,a)` | after |
|---|---|---|
| `1` vs `2` | `-1` / `1` | `-1` / `1` |
| `'a'` vs `'b'` | `-1` / `1` | `-1` / `1` |
| two `DateTimeImmutable`, different instants | `-1` / `1` | `-1` / `1` |
| `1` vs `1` | `0` / `0` | `0` / `0` |
| `1` vs `1.0` | `-1` / `-1` | `0` / `0` |
| `'1'` vs `'01'` | `-1` / `-1` | `0` / `0` |
| two `DateTimeImmutable`, same instant | `-1` / `-1` | `0` / `0` |

Every row that changes is one where the old comparator said "a first" and also "b first". `usort()` is not stable across such a comparator, so the resulting order depended on input order and on the sort implementation rather than on the ordering the user asked for — which is what the issue reports. There is no defined previous behaviour there to preserve, which is why I did not add a flag. Rows where the old comparator was well-defined are unchanged.

Happy to add an opt-in if you would still rather have one, but it would gate a code path that currently has no specified behaviour.

**Review comment.** Fixed — I dropped the comment entirely rather than rewording it; the three lines read fine on their own.

Sorry again about the leftover notes in the original description.

src/Expr/ClosureExpressionVisitor.php
dualfroz: Dropped the comment — the three lines stand on their own without it.

dualfroz:
ran phpstan the way ci does and it's clean, my bad. on the flag, this only changes the tie check, and the current one sorts differently depending on input order. can add a flag if you want though

